PR #4, opened 2018-04-11
change background color to default

tzing:
I found its background color rendered as the 'black' color I set in the terminal.
It results in poor readability.

screenshot (up: nvtop / bottom: htop)

this PR change the background color to the terminal default
screenshot - after the change

and it seems work on white background

I do not do the extra test since I don't have experience developing such the tools.
Please left comment if you have any concern.

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -292,12 +292,13 @@ void show_gpu_infos_ascii(
 
 static void initialize_colors(void) {
   start_color();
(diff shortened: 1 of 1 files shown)


Syllo:
First of all thanks for your contribution.

I've never had this issue as my configuration stays with the usual white on black background.
Now things will look beautifull for everyone :smile:

PR #10, opened 2018-10-09
fixed the path as a cmake arg

trimstray:
If you run `cmake ..` in **nvtop/build** directory you get:

```
[...]
mkdir -p nvtop/build && cd nvtop/build
cmake ..

CMake Error: The source directory "nvtop/nvtop" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
```

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -104,11 +104,11 @@ Two libraries are required:
 ```bash
 git clone https://github.com/Syllo/nvtop.git
 mkdir -p nvtop/build && cd nvtop/build
-cmake ..
+cmake ../..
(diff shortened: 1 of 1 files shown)


Syllo:
I'm not sure that there is a problem with these commands, if you strictly follow them in order.

It is true that if you change the directory to the git created nvtop folder before issuing the next commands it is not going to work because there will be a  ``nvtop/nvtop/build`` directory hierarchy.

trimstray:
You're right.

PR #25, opened 2018-11-30
--no-plot: remove required arg and document in usage

Khoyo:
The man page doesn't document the fact that an argument is required, and it seems more in line with other flags, such as `--no-cache`.

Also, document the option in usage (`--help`)


Syllo:
Thank you for the contribution.

I merged the first patch as 539ca24da9036a5ab0048ab43d0baf227d2cf237, the documentation had already been updated in 8e0edef3216b4068004400de0965816d1f22493f.

Khoyo:
Thanks !

Issue #33, opened 2019-02-09
switch quit to F10


frankenstein91:
I would very much appreciate it if you could end the application with F10. It would be the same as htop.

Issue #1, opened 2018-04-04
Cuda 9 Support 


Ly0n:
Thanks for this amazing work. I'm using this tool on a lot of Systems.
With Cuda 9 Installed on Ubuntu 16.04 I get the following problem:

```
-- The C compiler identification is GNU 5.4.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- No building type advertised, default to Release
CMake Error at /usr/share/cmake-3.5/Modules/FindPackageHandleStandardArgs.cmake:148 (message):
  Could NOT find NVML (missing: NVML_LIBRARIES) (found version "9")
Call Stack (most recent call first):
  /usr/share/cmake-3.5/Modules/FindPackageHandleStandardArgs.cmake:388 (_FPHSA_FAILURE_MESSAGE)
  cmake/modules/FindNVML.cmake:47 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:26 (find_package)
```

```
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 384.81                 Driver Version: 384.81                    |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  TITAN X (Pascal)    Off  | 00000000:82:00.0 Off |                  N/A |
| 23%   32C    P8    17W / 250W |    170MiB / 12188MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
|    0      1772      G   /usr/lib/xorg/Xorg                             6MiB |
|    0     18965      C   python                                       153MiB |
+-----------------------------------------------------------------------------+
```
Do you know how I can run nvtop with Cuda 9?

Syllo:
From the error message it managed to find the headers and extract the version (i.e. found version "9"). Although, the "NVML_LIBRARIES" variable, which should contain the path(s) to the library location, remains empty.

[...]

Issue #28, opened 2018-12-16
F3 or F2 is not closing the application


modem7:
F3 or F2 does not seem to be doing anything in this case. 

When I press either of them, nothing occurs. 

Fedora 25

keybraker:
You have to press **Q** from nvtop/src/nvtop.c line 338
`
case 'q':
        signal_exit = 1;
        break;
`

modem7:
Ah, that certainly works! Thanks!

Looks like the text needs to be modified 

![capture](https://user-images.githubusercontent.com/4349962/50060478-8df1e400-018c-11e9-8b4d-2f66d80ddefc.JPG)

Syllo:
Does the F1 or F2 work for the kill/sort menu? When the menu is active, F3 to quit is not available.
Which terminal emulator are you using?
Escape should also work to quit.

modem7:
Heya - sorry for the late reply!

Neither F1 nor F2 work as expected. 

Using Putty in this case

Syllo:
It seems like a problem with putty / terminal configuration then, see the [similar ubuntu bug](https://bugs.launchpad.net/ubuntu/+source/screen-profiles/+bug/317550).
Try setting the TERM environment variable as putty (`export TERM=putty`) as specified in the [ncurses faq](https://invisible-island.net/ncurses/ncurses.faq.html#xterm_generic).

aimuch:
The F1 is not available.
F1 will pop up the help menu.
I try modify `case KEY_F(1):`->`case KEY_F(9):`, and `case KEY_F(1):`->`case KEY_F(9):`, using F1 replace F1. The F9 works, however, the status bar is stall show F1.

Syllo:
Which help menu are you talking about?
In putty, running the executable as `TERM=putty nvtop` should fix the function key problem.

aimuch:
> Which help menu are you talking about?
> In putty, running the executable as `TERM=putty nvtop` should fix the function key problem.

![nvtop](https://user-images.githubusercontent.com/17581911/51108235-6a25ca80-182c-11e9-9532-3686c6ac7e8d.png)

Syllo:
With #34 and the configuration advice for putty in the `README` 765f355d16b364872d8c681a9b02b965a09688a3 this problem should be resolved.

PR #34, opened 2019-02-09
htop like control

frankenstein91:
I have fixed my problem https://github.com/Syllo/nvtop/issues/33

Sort is F6
Kill is F9
Exit is F10

Tested with ArchLinux 
Linux 4.20.6-arch1-1-ARCH #1 SMP PREEMPT Thu Jan 31 08:22:01 UTC 2019 x86_64 GNU/Linux
GPU: NVIDIA GeForce GTX 960

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -1407,9 +1407,14 @@ static void draw_options(struct nvtop_interface *interface) {
 }
 
 static const char *option_selection_hidden[] = {
-    "Kill",
     "Sort",
+    "Kill",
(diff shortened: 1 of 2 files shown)


Syllo:
Thanks you.
I think that this will also help with terminal emulator hijacking F1 #28 and putty bad configuration.

frankenstein91:
It will help for F1, but I know one where F10 does not work. But I think htop is well-known and F10 will be fixed.

PR #57, opened 2020-02-24
sorted signals

Red-Eyed:
Hello, just made PR that shows signals in right order
I've built it and run so it works.

- Made signals sorted.
- Made right correspondence  between signal value and name
- Refactored

image

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -35,6 +35,7 @@
 
 #define max(a, b) ((a) > (b) ? (a) : (b))
 #define min(a, b) ((a) < (b) ? (a) : (b))
+#define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))
 
 enum process_field {
   process_pid = 0,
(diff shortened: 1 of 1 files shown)


src/interface.c
```diff
@@ -1279,22 +1280,73 @@ static void draw_processes(struct device_info *dev_info,
...
 }
 
-static const char *signalsName[] = {
-    "Cancel",  "SIGABRT", "SIGALRM",   "SIGBUS",  "SIGCHLD", "SIGCONT",
-    "SIGFPE",  "SIGHUP",  "SIGILL",    "SIGINT",  "SIGKILL", "SIGPIPE",
-    "SIGQUIT", "SIGSEGV", "SIGSTOP",   "SIGTERM", "SIGTSTP", "SIGTTIN",
-    "SIGTTOU", "SIGUSR1", "SIGUSR2",   "SIGPOLL", "SIGPROF", "SIGSYS",
-    "SIGTRAP", "SIGURG",  "SIGVTALRM", "SIGXCPU", "SIGXFSZ",
+static const char *signalNames[] = {
+  "SIGHUP",
```
Syllo: Keep `"Cancel"` as the fist option please.
The code below assumes that the first of the list is cancel.

src/interface.c
```diff
@@ -1650,7 +1702,7 @@ static void option_do_kill(struct nvtop_interface *inter) {
   if (inter->process.option_window.selected_row == 0)
```
Syllo: This is the cancel-specific bit.

src/interface.c
Red-Eyed: Okay, but can I define `"Cancel"` as `SIGKILL` ?

src/interface.c
Red-Eyed: Oh, I got it. I thought that `"Cancel"` meant to kill process.

src/interface.c
Red-Eyed: Done

Red-Eyed:
So, I've added back "Cancel"

Syllo:
Thank you, merged as part of 1a985741e3b59d1cd2b335ac2c13c3b27f8d6988

PR #58, opened 2020-03-07
Update Arch Linux installation instructions

felixonmars:
It has been added to [community] repository: https://www.archlinux.org/packages/community/x86_64/nvtop/


Syllo:
Thanks for the patch and the inclusion in the community repository

PR #87, opened 2021-01-04
Add additional path hints for nvml

ebetica:
Some systems (like SLURM) set CUDA_HOME, which we can use to find NVML.


Syllo:
Hello,

Thanks for your contribution, it is merge as of 4e981a749d87759eb5bcecd39493c9c9b4b9b5d1 with some slight modifications.

PR #91, opened 2021-01-27
Add columns for per-process GPU, Encoder, and Decoder usage

philipl:
These stats exist, although they appear to behave strangely. They seem
to take a significant amount of time to rise from zero to the correct
value, although they do seem to eventually converge.

### DIFF
--- a/include/nvtop/extract_gpuinfo.h
+++ b/include/nvtop/extract_gpuinfo.h
@@ -40,6 +40,9 @@ struct gpu_process {
   size_t cpu_memory_virt;
   size_t cpu_memory_res;
   double cpu_usage;
+  double gpu_usage;
+  double enc_usage;
+  double dec_usage;
 };
 
 enum dev_info_valid {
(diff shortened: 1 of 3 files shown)


Syllo:
Hey,

Nice to see this being investigated by someone.
And with a pull request on top of that, thanks!

As you pointed out, the stat doesn't seem to reflect the current state of the GPU, but seems to be lacking behind.

I'll play with it to see if the latency can be reduced.
If not I might include it behind a command line option with a warning about the latency.

Syllo:
Hi,

I found out why the values were lagging behind.
The function ``nvmlDeviceGetProcessUtilization`` takes as last argument a ``lastSeenTimeStamp`` parameter which needs to be stored from a previous call to get the utilization since the last call. Otherwise the function returns the utilization over a larger period of time which depends on the driver.

I did a big overhaul of the internals recently and added the per-process utilization with the fix at the end.

Thanks for pointing this out

Take care

PR #98, opened 2021-03-22
Fix: initialize_device_info failed cause double free

oToToT:
When initialize_device_info failed, it will free allocated memory and return 0. However, the caller might free the memory again and cause a double free problem.


Syllo:
I see, this case was not handled properly.
Thanks for the fix.
Take care!

PR #103, opened 2021-04-28
More paths for NVML library

ebetica:
Unfortunately, on a new system, I am once again unable to find the nvml libraries. Here's where you can find it.


Syllo:
Yep, the header seems to be, either absent because of not being installed with the cuda package, or put at a custom location!
Thanks for the contribution.
Take care!

Rebased in 84902667d4d9cf210f1461eb60b504c6ed1f2cfd

PR #104, opened 2021-05-04
Fix: handle empty `/proc/[PID]/cmdline` properly

oToToT:
In some rare cases, the file `/proc/[PID]/cmdline` might be empty. This causes nvtop display wrong information about command being executed or even leak some data inside current memory.

For example, it might leak something similar to `/etc/passwd`.
image

I'm not sure how htop or ps handle this kind of case, but I think it's OK to display nothing in command section as it's actually empty there.

### DIFF
--- a/src/get_process_info_linux.c
+++ b/src/get_process_info_linux.c
@@ -96,6 +96,10 @@ void get_command_from_pid(pid_t pid, char **buffer) {
   }
   fclose(pid_file);
 
+  if (total_read == 0) {
+    (*buffer)[0] = '\0';
+  }
+
   for (size_t i = 0; total_read && i < total_read - 1; ++i) {
     if ((*buffer)[i] == '\0')
       (*buffer)[i] = ' ';


oToToT:
I'm not very sure that whether `/proc/[PID]/cmdline` will always contain '\0' in the end or not. Might it be much better to append a '\0' directly to the end of the buffer?

Syllo:
Hi,
Indeed, a zero character read can be returned for this file.
Could you please check that the fix in the branch [cmdline_fix](https://github.com/Syllo/nvtop/commit/b1164c8ec07b47fc81afebdcc3c416064cd8ded9) resolves this issue?

oToToT:
> Hi,
> Indeed, a zero character read can be returned for this file.
> Could you please check that the fix in the branch [cmdline_fix](https://github.com/Syllo/nvtop/commit/b1164c8ec07b47fc81afebdcc3c416064cd8ded9) resolves this issue?

Yes, this do resolve this issue and I think it's a better fix than mine.

Syllo:
Merged into master, you're welcome.
Take care!

PR #119, opened 2021-09-08
A spurious fprintf Is breaking layout 

gravicle:
Please see "Comparing GPU rates" appearing all over the term GUI in the screenshot below.

ScreenCapture-2021-09-08-15 58 20-iTerm


Syllo:
Thanks for the fix.
I added that by mistake after a debug session ...

PR #120, opened 2021-09-09
Update Dockerfile and build instructions

XuehaiPan:
Changes:

1. combine `apt update` and `apt install` commands into a single statement (as suggested in the document [Best practices for writing Dockerfiles](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run)).
2. remove build-cache after a successful build.
3. update build instructions.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,1 @@
-*.o
-*ctags
-build/
-cmake-build*/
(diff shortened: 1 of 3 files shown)


Syllo:
Thank you @XuehaiPan.

PR #134, opened 2022-03-10
User utility along with non-privileged user installation script

neverloses87:
I packed the dependency within the repo and provided the building script for non-root users.

Moreover, I added the user utility function so that users could select whom to be monitored in the terminal.

The user utility is built on top of the original code with slight modification on function interface and some additional stuff.


Syllo:
Hey @jeff999955,

So sorry I saw your pull request some time ago but forgot to follow up.

I won't merge the ncurses source because I don't think that it belongs in this repository, I'll probably add a link to the release page instead.
I will gladly cherry pick the bits for sorting/filtering by user name and the install instruction for a user without system install permissions.

Thank you for contributing to nvtop.

Issue #106, opened 2021-05-21
Support for AMD and Intel devices


Syllo:
I am opening this up because of [this thread of discussion on the intel-sgx mailing list](https://lists.freedesktop.org/archives/intel-gfx/2021-May/266057.html).

It seems that the GPU utilization for AMD and Intel will be available in future kernels, at least for the processes spawned by the user itself.

Marking this as a possible improvement once the support lands into the mainline kernel.

thesunexpress:
Might be a better idea to just call it ' gputop ' instead then...

el1-x:
@Syllo any update on this?

Syllo:
I will keep this issue around as a meta-issue to centralize the information on other vendors / driver support.

thomas-zimmerman:
I'm on a 5.17.4 kernel with a RX 6700 XT AMD GPU and the output that I get is:
```
No GPU to monitor.
```
This is on Ubuntu 22.04 with `nvtop_1.2.2-1_amd64.deb`. I built from source and I get the same output. Any thing I should run to get output on what's missing on this kernel to get output?

Syllo:
@thomas-zimmerman I created #144 for your issue

Syllo:
The Intel sysfs change has been proposed for Linux 5.19: https://lore.kernel.org/dri-devel/Ymkfy8FjsG2JrodK@tursulin-mobl2/

servesh:
@Syllo Would there be any interest in adding support for Intel GPUs via level zero Sysman interface (https://spec.oneapi.io/level-zero/latest/sysman/PROG.html)?

Syllo:
@servesh Thanks for the link. I was not aware that oneAPI provided such library.
This seems like a good opportunity to simplify the accelerator discovery and query.
I'll see if I can make this work.

servesh:
@Syllo Sure, Iet me know once you have something that I can test and take a look. Happy to contribute to the code base as well.

HTOgit:
@Syllo Is it possible to show RX/TX on AMD GPUs? I'm on Linux 5.19.

Syllo:
@HTOgit
The support for the [pcie_bw file](https://dri.freedesktop.org/docs/drm/gpu/amdgpu.html#pcie-bw) is in place.
However this file is empty on the system I am testing (RX 6800XT, Kernel 5.19.6).

The only explanation that I see is that the support has not been implemented in the driver:
[...]

PR #135, opened 2022-03-21
AMDGPU support

zhuyifei1999:
A picture is worth a thousand words:
Screenshot_2022-03-21_10-53-19 c

This PR isn't fully ready for merge yet. I think I'd like some feedback and do some polishing first (hence the RFC). My AMDGPU is also an iGPU / APU which limits how much I can test myself.

Most the access is using libdrm and kernel APIs. <drm/amdgpu_drm.h> is a kernel UAPI header so I directly included it. For libdrm I'm using dlopen and re-declaring the constants to avoid a compile-time dependency, just like what the original code does with NVML.

I also did some code refactoring to make my life easier, hope that is okay. ~~I really cannot deal with typedef structs~~

CC #106, i915 kernel changes doesn't seem to have merged into mainline yet.


Syllo:
Amazing!
I will try to take a look at it this weekend.

Syllo:
I successfully tested on a system I have access to with an AMD Radeon RX 6800 XT.
I also fixed some issues I encountered while browsing the code.

For the fan speed, I saw that you added a comment. I did not find a way to get it through AMDGPU. lm_sensors seem to get the correct info though, so I will investigate where to retrieve it.

zhuyifei1999:
Awesome!

Yeah I didn't bother with checking for `close(-1)` since it's practically a no-op, similar to `free(NULL)`; though unlike free it has syscall overhead.

The reason I didn't do fan speed was that it seemed like it was slightly convoluted to retrieve. Rather than being a libdrm API call, I have to go through sysfs; doable but code can get messy and I can't test it.

src/extract_gpuinfo_amdgpu.c
```diff
@@ -301,6 +309,11 @@ static void gpuinfo_amdgpu_shutdown(void) {
   struct gpu_info_amdgpu *allocated, *tmp;
 
   list_for_each_entry_safe(allocated, tmp, &allocations, allocate_list) {
+    if (allocated->hwmonFD >= 0)
+      close(allocated->hwmonFD);
+    if (allocated->sysfsFD >= 0)
+      close(allocated->sysfsFD);
+    fclose(allocated->fanSpeedFILE);
```
zhuyifei1999, line 316: Hmm, I'm actually relying on the closing of the program to free all the resource usages; it doesn't actually go through the device list to free each device's handles. This allocate_list is more of freeing the array of devices itself.

How I made it work is pre-allocating an array of devices, and when we successfuly get the handles for one device, add that device to the exported linked list. The whole array is added the "allocations" linked list once.

Should we add this free-each-device's handles?

src/extract_gpuinfo_amdgpu.c
```diff
@@ -352,6 +365,46 @@ static void authenticate_drm(int fd) {
...
+    int hwmondirFD = openat(gpu_info->sysfsFD, hwmon, O_RDONLY);
+    if (hwmondirFD >= 0) {
+      DIR *hwmonDir = fdopendir(hwmondirFD);
+      if (hwmonDir) {
+        struct dirent *dirEntry;
+        while ((dirEntry = readdir(hwmonDir))) {
+          // There should be one directory inside hwmon, with a name having the following pattern hwmon[0-9]+
+          if (dirEntry->d_type == DT_DIR) {
+            size_t matchLen = 0;
+            for (matchLen = 0; matchLen < sizeof(hwmon) - 1; ++matchLen) {
```
zhuyifei1999, line 386: Could be strncmp(dirEntry->d_name, "hwmon", sizeof("hwmon") - 1)?

src/extract_gpuinfo_amdgpu.c
```diff
@@ -352,6 +365,46 @@ static void authenticate_drm(int fd) {
...
+            }
+            // We found our candidate
+            if (matchLen == sizeof(hwmon) - 1)
+              break;
+          }
+        }
+        if (dirEntry) {
+          gpu_info->hwmonFD = openat(dirfd(hwmonDir), dirEntry->d_name, O_RDONLY);
+        } else
+          gpu_info->hwmonFD = -1;
```
zhuyifei1999, line 398: How about instead of adding these else-s, pre-initialize `gpu_info->hwmonFD = -1;` and only assign it if things go well? Could also avoid nested ifs by returning early.

src/extract_gpuinfo_amdgpu.c
Syllo: I see. Why are you using a list if there is only one element in the "allocations" list, which is static to this file? 

I guess that relying on the program termination to close the file descriptors is fine in that case.

src/extract_gpuinfo_amdgpu.c
Syllo: Correct, I'll fix that.

src/extract_gpuinfo_amdgpu.c
Syllo: You are right.
I do the convoluted if/else spaghetti as a first pass when the code is still changing many times.
I will clean this up.

src/extract_gpuinfo_amdgpu.c
zhuyifei1999: Yeah, at first I wasn't sure how many times `get_device_handles` gets called :)

Syllo:
Other than that, does the patch work on your APU?
I am wondering, because in that case the fan is shared with the CPU and might not be registered with HWMON as such.

zhuyifei1999:
The APU does not expose fan information on the GPU's hwmon, AFAICT.

```
zhuyifei1999@zhuyifei1999-ThinkPad-P14s-Gen-2a /sys/bus/pci/devices/0000:08:00.0 $ lspci | grep VGA
08:00.0 VGA compatible controller: Advanced Micro Devices, Inc. [AMD/ATI] Cezanne (rev d1)
zhuyifei1999@zhuyifei1999-ThinkPad-P14s-Gen-2a /sys/bus/pci/devices/0000:08:00.0 $ ls
ari_enabled                devspec           hwmon          max_link_speed           msi_irqs                           pp_dpm_sclk            rescan         smartshift_apu_power
backlight                  dma_mask_bits     i2c-0          max_link_width           numa_node                          pp_dpm_socclk          reset          smartshift_dgpu_power
boot_vga                   driver            i2c-1          mem_info_gtt_total       pcie_replay_count                  pp_force_state         reset_method   subsystem
broken_parity_status       driver_override   i2c-12         mem_info_gtt_used        power                              pp_mclk_od             resource       subsystem_device
class                      drm               i2c-2          mem_info_preempt_used    power_dpm_force_performance_level  pp_num_states          resource0      subsystem_vendor
config                     enable            i2c-3          mem_info_vis_vram_total  power_dpm_state                    pp_od_clk_voltage      resource0_wc   thermal_throttling_logging
consistent_dma_mask_bits   firmware_node     iommu          mem_info_vis_vram_used   power_state                        pp_power_profile_mode  resource2      uevent
consumer:pci:0000:08:00.1  fw_version        iommu_group    mem_info_vram_total      pp_cur_state                       pp_sclk_od             resource2_wc   vbios_version
current_link_speed         gpu_busy_percent  irq            mem_info_vram_used       pp_dpm_dcefclk                     pp_table               resource4      vendor
current_link_width         gpu_metrics       link           mem_info_vram_vendor     pp_dpm_fclk                        product_name           resource5
d3cold_allowed             graphics          local_cpulist  modalias                 pp_dpm_mclk                        product_number         revision
device                     hdcp_srm          local_cpus     msi_bus                  pp_dpm_pcie                        remove                 serial_number
zhuyifei1999@zhuyifei1999-ThinkPad-P14s-Gen-2a /sys/bus/pci/devices/0000:08:00.0 $ ls hwmon/
hwmon1
zhuyifei1999@zhuyifei1999-ThinkPad-P14s-Gen-2a /sys/bus/pci/devices/0000:08:00.0 $ ls hwmon/hwmon1/
device  freq1_input  freq1_label  in0_input  in0_label  in1_input  in1_label  name  power  power1_average  power1_label  subsystem  temp1_input  temp1_label  uevent
```

Syllo:
I think that we are almost good.
I will update the README and check for the amdgpu_drm.h to conditionally compile the support for AMDGPU.

zhuyifei1999:
Nice!

I'm including amdgpu_drm.h directly because it's a [kernel header](https://github.com/torvalds/linux/blob/master/include/uapi/drm/amdgpu_drm.h), and I think any sane compiler toolchain on Linux would include the kernel headers. Though, up to you, could copy the declarations too since kernel ABI has guaranteed stability.

Syllo:
The problem with NVIDIA was that the nvml header was not packaged on all the distributions, and downloading a recent header could break the build. Also I was not certain I could redistribute a version of the nvidia header. That is why I included the definitions directly.

Things are different with libdrm. The header are consistently available on any distribution and updated with the kernel. Much cleaner than NVIDIA.

Hence, relying on libdrm for AMDGPU support is fine in my opinion!

If you are fine with the latest changes I will merge it into master.

zhuyifei1999:
> The problem with NVIDIA was that the nvml header was not packaged on all the distributions, and downloading a recent header could break the build. Also I was not certain I could redistribute a version of the nvidia header. That is why I included the definitions directly.

I see.

> Neat Videocard TOP

Nice naming lol

> If you are fine with the latest changes I will merge it into master.

No objections. Go ahead. Thanks!

PR #153, opened 2022-06-13
removed some warning messages

brccabral:
Hi,
When compiling I was getting some warning messages due to printf formatting, so I just changed the formatting to remove the warnings.


Syllo:
Thanks a lot @brccabral for your patch.

PR #154, opened 2022-06-18
Link to rt on Linux

hmaarrfk:
I guess I'm trying to build on some platforms where `rt` isn't linked to automatically. 

I think this is the correct way to get things to work.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -63,6 +63,9 @@ target_compile_definitions(nvtop PRIVATE _GNU_SOURCE)
 
 target_link_libraries(nvtop
   PRIVATE ncurses m ${CMAKE_DL_LIBS})
+if(UNIX AND NOT APPLE)
+    target_link_libraries(nvtop PRIVATE rt)
+endif()
 
 install (TARGETS nvtop
   RUNTIME DESTINATION bin)


hmaarrfk:
xref: https://github.com/conda-forge/staged-recipes/pull/19330

Syllo:
Hello @hmaarrfk,

Could you please elaborate on what symbol from librt is necessary for nvtop?
I don't seem to be needing it on my workstation, but it might be needed on others.

hmaarrfk:
Here is a fuller log.

I guess from what I've seen, certain distributions will link with rt by default.

However, others still ask for it to be explicigely stated.


This is the error message that appears without the proposed patch

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=550926&view=logs&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=e5c8ab1d-8ff9-5cae-b332-e15ae582ed2d&l=306

hmaarrfk:
See build process here:

https://github.com/conda-forge/nvtop-feedstock/pull/2

zhuyifei1999:
So that symbol is `clock_gettime`

hmaarrfk:
I guess I is an old version of glibc. I forget if 2.17 or 2.12.

hmaarrfk:
Hmm thank you for educating me. 

It seems that by default we do indeed compile for 2.12 

There seem to be some disadvantages to linking to rt including the implicit dependency and usage of pthreads.

I updated to 2.17 instead which i think is warranted for this particular package.

PR #156, opened 2022-08-08
add gcc-c++ to Fedora instructions

Seibz:
Making note that you'll the library gcc-c++ is required for cmake to compile this application, even if the "Development  Tools" group is installed on Fedora 36.


Syllo:
@Seibz looks good tank you. This may be a consequence of introducing tests in c++.

PR #160, opened 2022-08-26
Fix format-security warnings

felixonmars:
nvtop 2.0.3 fails to build on Arch Linux with `-Werror=format-security` enabled. Let's fix it.


Syllo:
LGTM. Is that blocking for Arch or is that a local cflag?

felixonmars:
It's a globally enabled CFLAG for Arch. I have added this patch already, for now :)

Syllo:
All right, thanks for the patch and the heads up.
I've added the flag to the debug build to catch similar issues in the future.

felixonmars:
That's awesome. Thanks!

PR #166, opened 2022-09-21
Add snap packaging and install instructions

Nightmayr:
I've added a `snapcraft.yaml`, which builds a snap successfully allowing nvtop to be installed on any Linux distro with snapd installed and be more up to date (as well as its dependencies) than the version that is usually in the distribution repos. Testing the snap locally on my machine shows it working as expected.

To transfer the snap into your ownership you would need to:

1. create a developer account on [Snapcraft.io](https://snapcraft.io/) 
2. I would then add your account as a collaborator on the snap
3. Raise a request on the Snapcraft forum to transfer the snap to you (I can stay on as a collaborator if you wish)

You can then configure in the developer dashboard to have automated builds of the snap when a push is made to the default branch of this repo. The resulting snap will end up on the edge channel, from which it can be promoted to stable on the dashboard after manual testing.

The snap currently needs the `hardware-observe` interface to be connected for AMD GPU fan speeds to be visible and `process-control` to kill processes. I've raised an auto-connect [request](https://forum.snapcraft.io/t/autoconnect-hardware-observe-for-nvtop/31869) on the forum which should hopefully be approved, meaning that on install the snap won't need the user to manually enable the permission.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -173,6 +173,8 @@ Several libraries are required in order for NVTOP to display GPU information:
   snap install nvtop
   snap connect nvtop:hardware-observe
   snap connect nvtop:process-control
+  snap connect nvtop:system-observe
+  snap connect nvtop:kubernetes-support
   ```
 
 ### Docker
--- a/README.markdown
+++ b/README.markdown
@@ -167,6 +167,13 @@ Several libraries are required in order for NVTOP to display GPU information:
   sudo layman -a guru && sudo emerge -av nvtop
   ```
 
+### Snap
+
+- ```bash
+  snap install nvtop
+  snap connect nvtop:hardware-observe
+  ```
+
 ### Docker
 
 - NVIDIA drivers (same as above)
--- a/README.markdown
+++ b/README.markdown
@@ -172,6 +172,7 @@ Several libraries are required in order for NVTOP to display GPU information:
 - ```bash
   snap install nvtop
   snap connect nvtop:hardware-observe
+  snap connect nvtop:process-control
   ```
 
 ### Docker
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,72 @@
+name: nvtop
+base: core22
+adopt-info: nvtop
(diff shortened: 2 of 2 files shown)


Syllo:
Helo @Nightmayr,
That looks fantastic, thank you.

So that should work with Ubuntu >= 20.04 since they ship snap by default from this release on if I am correct.

Did you try with an NVIDIA GPU, or AMD only thus far?

Nightmayr:
Hi @Syllo, in terms of Ubuntu the snap should work with Ubuntu >= 18.04 (oldest supported LTS with snap enabled by default). I believe it should even work with unsupported Ubuntu versions going back to 16.04.

I have tried the snap on my local system which contains both an Nvidia 2080ti and AMD RX580. Nvtop shows both GPUs, but does require some additional permissions for the processes of AMD GPUs to be visible as otherwise the sandboxing shows apparmor denials like:

```
= AppArmor =
Time: Sep 21 22:11:56
Log: apparmor="DENIED" operation="open" profile="snap.nvtop.nvtop" name="/proc/5962/fdinfo/" pid=147463 comm="nvtop" requested_mask="r" denied_mask="r" fsuid=1000 ouid=1000
File: /proc/5962/fdinfo/ (read)
```

A user would need to connect the following interfaces allowing nvtop to read that path:
```
snap connect nvtop:system-observe
snap connect nvtop:kubernetes-support
```

`kubernetes-support` doesn't make sense for nvtop, but it's currently the only interface with an [apparmor rule](https://github.com/snapcore/snapd/blob/fc72b84550e78ff0e8f4398e222080fa33d7a96b/interfaces/builtin/kubernetes_support.go#L158) allowing access to `@{PROC}/@{pid}/fdinfo/`. In the auto-connect request I raised in the forum it looks like the apparmor rule will either be added to the _"base template for all snaps, or at least in the system-observe interface"_. Once that happens I'll remove the `kubernetes-support` permission. On distributions without apparmor like Fedora, the snap runs with 'partial confinement' and should work without needing any permissions to be set.

Issue #173, opened 2022-10-25
Snap installed nvtop not operational


Dhruval360:
I had installed nvtop using snap and it stopped working with the following error since the past few days.

```
cannot update snap namespace: cannot create symlink in "/usr/share/libdrm": existing file in the way
snap-update-ns failed with code 1
```

When I checked what /usr/share/libdrm had, I found a file named amdgpu.ids with the following content:

```
# List of AMDGPU IDs
#
# Syntax:
# device_id,	revision_id,	product_name        <-- single tab after comma

1.0.0
15DD,	C3,	AMD Radeon(TM) Vega 3 Graphics
15DD,	CB,	AMD Radeon(TM) Vega 3 Graphics
15DD,	CE,	AMD Radeon(TM) Vega 3 Graphics
15DD,	D8,	AMD Radeon(TM) Vega 3 Graphics
15DD,	CC,	AMD Radeon(TM) Vega 6 Graphics
15DD,	D9,	AMD Radeon(TM) Vega 6 Graphics
15DD,	C2,	AMD Radeon(TM) Vega 8 Graphics
15DD,	C4,	AMD Radeon(TM) Vega 8 Graphics
15DD,	C8,	AMD Radeon(TM) Vega 8 Graphics
15DD,	CA,	AMD Radeon(TM) Vega 8 Graphics
15DD,	D1,	AMD Radeon(TM) Vega 8 Graphics
15DD,	D5,	AMD Radeon(TM) Vega 8 Graphics
15DD,	D7,	AMD Radeon(TM) Vega 8 Graphics
15DD,	C3,	AMD Radeon(TM) Vega 10 Graphics
15DD,	D0,	AMD Radeon(TM) Vega 10 Graphics
15DD,	C1,	AMD Radeon(TM) Vega 11 Graphics
15DD,	C6,	AMD Radeon(TM) Vega 11 Graphics
15DD,	C9,	AMD Radeon(TM) Vega 11 Graphics
15DD,	D3,	AMD Radeon(TM) Vega 11 Graphics
15DD,	D6,	AMD Radeon(TM) Vega 11 Graphics
15DD,	81,	AMD Ryzen Embedded V1807B with Radeon Vega Gfx
15DD,	82,	AMD Ryzen Embedded V1756B with Radeon Vega Gfx
15DD,	83,	AMD Ryzen Embedded V1605B with Radeon Vega Gfx
[...]

PR #176, opened 2022-11-06
snap: disabling libdrm amdgpu.ids conflicting symlink, enabling intel

Nightmayr:
Fixes https://github.com/Syllo/nvtop/issues/173

This amounts to a quick fix for now just to get the snap working again, but in the background I'll carry on trying to get an answer on how to use the nvtop snap's amdgpu.ids over the core22 provided file. This change also enables nvtop to build with intel support enabled.


Syllo:
Thanks. I promoted the build based on this patch as the latest release version.
I am currently not using libdrm_intel.so but may be using it to retrieve the memory usage if it exposes that. We can keep the extra dependency for now.

PR #188, opened 2023-02-05
fix multiple gpus selection

cfwen:
since line 73 has moved options backward by one space (towards head) when doNotMonitor is true, idx shouldn't increase in such case. 

For example, we have two gpus with [1:doNotMonitor:false, 2:doNotMonitor:true], current code would first check gpu1 and move it to the tail, after that, gpus' options would become [2:doNotMonitor:true, 1:doNotMonitor:false]. Then in next iteration, code will check gpu1 again and gpu2's option is ignored.

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -1788,14 +1788,17 @@ void interface_key(int keyId, struct nvtop_interface *interface) {
       interface->process.option_window.selected_row = 0;
     }
     break;
+  case 'l':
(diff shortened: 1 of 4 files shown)


Syllo:
You are right, thanks.

That wasn't the only error in this function: in the second loop idx was decremented instead of the inner loop induction variable!

PR #190, opened 2023-02-17
Implement a lookup table to identify newer AMD GPUs

paklui:
We have implemented some changes to help identifying some of the newer AMD Instinct GPUs that we work on. The issue is that some of the newer GPUs were getting identified, because libdrm that we relied on to get the GPU names which usually comes from OS but most of the popular OS distros used today didn’t have the latest list. We implemented a look up table in case the GPU weren’t being identified for the AMD GPUs.


zhuyifei1999:
(Nitpick) I'm wondering, can these IDs be pre-parsed at, say, build time rather than at run time?

amdsc21:
Hi @zhuyifei1999, I checked the libdrm. It has been implemented the same way, as it reads amdgpu.ids at run time. The advantage is that it searches the database (IDs), and it supports all listed GPUs in the database. Also, the marketing name check is put in static info, which should be checked once. By putting the marketing name checking at run time, it allows us to compile once and supports all listed GPUs on a LinuxDistro. What do you think? Thanks.

amdsc21:
Hi @zhuyifei1999, thanks for the advice. We choose to put the hardware info in a header file, so when nvtop is distributed, we don't need to copy the database file, amdgpu.ids, which is also hardcoded and has to be updated after new GPU release. As a Linux Distro may choose old version libdrm, it may not get correct marketing name for latest GPUs. Some old Linux Distro may not update libdrm, and in this way, marketing name is always wrong for new GPUs. This pull request is hoping to by-pass Linux kernel and Distro to provide a much quicker way to support new GPUs.

If you look at the driver libdrm source code, it parses each line too, which has memory allocation, and string manipulations. If you have concern about memory corruption / leak, one easiest modification is to provide a char array as buffer to libdrm and this pull request. I would not worry about performance, as the marketing name query happens once for each GPU, and our computer can parse millions of lines per second. :-)

Thanks again and we will send another pull request after changes.

Syllo:
Hello,
Thanks for posting that after our offline discussion.
Sorry I have little time due to moving recently. I'll take a deeper look tomorrow but so far this looks reasonably good.

Syllo (approved):

PR #194, opened 2023-02-25
add Fedora, RHEL, CentOS distribution packages

Mormacill:
Since a few month nvtop is finally provided as [distribution package](https://src.fedoraproject.org/rpms/nvtop) for Fedora.
It is also built for EPEL which allows using nvtop on Red Hat EL 8 and 9 as well as clones like Rocky Linux and AlmaLinux.
I would appreciate it if you could include these changes in the readme to draw attention to the distro builds.

(I also updated the link to the Ubuntu section since it was broken)


Syllo:
Thank you @Mormacill.
Nice to see it being picked up by distros.

PR #224, opened 2023-07-20
MSM: Fixes memory usage

Sonicadvance1:
Using `sysinfo` originally was a mistake. The `freeram` member that sysinfo doesn't account for the difference between used, buffers, shared, and cache.

Instead switch over to processing `/proc/meminfo` directly and parse the `MemAvailable` and `MemTotal` keys.

`MemTotal` gives us the same information back as sysinfo's totalram, so that doesn't change.

`MemAvailable` instead returns the amount of memory available for userspace allocations without causing swapping or OOM. Not accounting for memory currently live from caching.

This still isn't ideal since this might not be what the user expects, but it is definitely better than the memory usage always showing that there's only a few hundred megs free from cache occupying all the RAM.

To be more ideal, we'll likely need to introduce new concepts for how to display RAM usage on UMA systems. Likely separating GPU buffers, application allocations, and external applications consuming system RAM.

That's more involved and I'm not sure how desirable that will be.

### DIFF
--- a/src/extract_gpuinfo_msm.c
+++ b/src/extract_gpuinfo_msm.c
@@ -95,6 +95,7 @@ unsigned msm_gpu_count;
 static struct gpu_info_msm *gpu_infos;
 
 static void *libdrm_handle;
+static FILE* meminfo_file = NULL;
 
 static int last_libdrm_return_status = 0;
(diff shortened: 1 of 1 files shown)


Syllo (approved):
Looks good to me. Thanks

PR #226, opened 2023-07-24
Corrected proper nouns

comradekingu:
Came to check the spelling for use elsewhere.


Syllo:
Hello,
Thanks for your patch.
The only change I am not agreeing with is the `information` -> `info` change. Could you revert that please?

PR #253, opened 2023-11-16
cmake version upgrade docker script. 3.10 to 3.18

Jun0922:
docker image was not built because of the lower version of cmake.
I made some cmake updating script on the Dockerfile so that the image is now available without any error.
Please check before/after below and approve code so that everyone can use docker image of this awesome nvtop.

Thanks for your great contribution! :)

- before
image

- after
- build
image

- running container
image


Syllo:
Thank you for noticing this and fixing it :wink:

PR #260, opened 2023-12-27
Updated to fix errors in build to run on M2 Air

SamRaymond:
Updated file to include missing functions to run on Macbook Air M2, tested, running well on mac.


Syllo:
Cheers, I added that recently for the Linux version and did not port it to Mac

PR #263, opened 2024-01-09
MSM: Add support for a750 and ignore speed-bin

Sonicadvance1:
Fairly basic. Adds three new IDs for Adreno 750 products and if the full GPUID isn't found on first linear search, mask off the speed-bin bits of the GPUID and try again.


Syllo:
Cheers

PR #272, opened 2024-02-23
Remove import of kcmp

hmaarrfk:
It seems this kcmp is longer used.

Let me know if there is anything else you need.


Syllo:
Good catch. I refactored the code that was there in the [shared fdinfo extraction function a while back](https://github.com/Syllo/nvtop/blob/master/src/extract_processinfo_fdinfo.c#L165).

Cheers

PR #278, opened 2024-03-06
Fix icon in appstream

michel-slm:
When validating the AppStream metadata as part of the Fedora packaging, we noticed it failed without this applied


Syllo:
Cheers

PR #282, opened 2024-03-12
Add missing source file dependency for interfaceTests

larunbe:
Otherwise, linker will complain about processinfo_enable_disable_callback_for being left undefined.


Syllo:
Cheers

PR #283, opened 2024-03-14
panfrost: replace debugfs profiling knob with sysfs

larunbe:
Kernel commit b12f3ea7c188 ("drm/panfrost: Replace fdinfo's profiling debugfs knob with sysfs") did away with the debugfs knob that lets profilers enable the hardware support for job accounting and fdinfo numbers. It was replaced with a similar interface, but this time around through sysfs.

Reflect this change in nvtop.

### DIFF
--- a/src/extract_gpuinfo_mali_common.c
+++ b/src/extract_gpuinfo_mali_common.c
@@ -94,6 +94,10 @@ bool mali_init_drm_funcs(struct drmFuncTable *drmFuncs,
   if (!drmFuncs->drmCommandWriteRead)
     goto init_error_clean_exit;
 
+  drmFuncs->drmGetDeviceFromDevId = dlsym(state->libdrm_handle, "drmGetDeviceFromDevId");
+  if (!drmFuncs->drmGetDeviceFromDevId)
+    goto init_error_clean_exit;
+
   drmFuncs->drmIoctl = dlsym(state->libdrm_handle, "drmIoctl");
   if (!drmFuncs->drmCommandWriteRead)
     goto init_error_clean_exit;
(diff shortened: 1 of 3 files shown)


warpme:
@larunbe 
thx for this fix!
It works well for me on 6.10 mainlie on g52 (e.g. rk3566) and g610 (rk3588). Unfortunatelly it not fixing non-working nvtop on g31 (e.g. h616).
On g31 i'm getting:
```
Profiling state not available in Panfrost
No GPU to monitor. 
```
Maybe you have idea where issue might be?

Syllo:
Hi,
Thanks for the patch.
If I understand correctly this patch will make nvtop non working on kernels < 6.10 since it's the old interface or has your patch been back-ported to earlier kernel versions?

larunbe:
> Hi, Thanks for the patch. If I understand correctly this patch will make nvtop non working on kernels < 6.10 since it's the old interface or has your patch been back-ported to earlier kernel versions?

Yeah, that's right, 6.10 kernel release has the change that replaces the debugfs interface with sysfs.

chealy:
I think this patch really needs to be enhanced to allow for running nvtop both with older (debugfs) and newer (sysfs) kernels as we cannot assure we will have the right version of nvtop paired with the kernel.

Syllo:
This would be better.
There is already an existing function to get the Linux kernel version that could be reused to achieve this.

The question is: Are there enough users out there with an old kernel to warrant supporting both?
The new interface seems the way forward and the old one seems to have been quire ephemeral imho. Maybe we can direct old kernels users to nvtop v1.1.0 instead?

PR #293, opened 2024-05-03
Update amdgpu_ids.h for RX 7900 XTX / XT

numas:
Add ids for RX 7900 XTX + RX 7900 XT, may work for other RX 7xxx series


numas:
I didn't check a clean build beforehand, and this change is not needed.

Issue #294, opened 2024-05-06
Option for Showing Only Compute/Graphical


stellarpower:
Hi,

Thanks for a great tool!

I am only using nvtop when I'm checking use for neural network training, so for me, the graphical processes can clutter the display (and I often need to watch side-by-side with other windows to keep an eye on what is using up resources).

Would it be possible to have an option in the settings to display only graphical or only compute processes, depending on what the user is interested in most?

Thanks

PR #295, opened 2024-05-13
Hide process list

ring-c:
CLI option to hide process list
Related #294 #191


Syllo:
Thanks you for adding the cli option. I'll add the interface option so that we can merge that.

PR #298, opened 2024-05-25
Make the appimage be able to work with libfuse3

Samueru-sama:
It uses this version of appimagetool instead because the built in one of linuxdeploy still depends on fuse2: https://github.com/probonopd/go-appimage

Note that I went and also did the --appimage-extract for the appimagetool like you were doing it with linuxdeploy, I don't think it is needed anymore because that appimagetool itself also works with fuse3, but I didn't test the script with podman so I can't be 100% sure. 

One potential breaking change is that the appimage will have a different name, because this version of appimagetool requires one to provide the version of the app to add it to the title of the resulting AppImage.

### DIFF
--- a/AppImage/make_appimage.sh
+++ b/AppImage/make_appimage.sh
@@ -23,6 +23,14 @@ get_linuxdeploy() {
   ./linuxdeploy-x86_64.AppImage --appimage-extract
 }
 
+get_appimagetool() {
+	ARCH=x86_64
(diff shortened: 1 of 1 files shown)


Syllo:
Hello,
Just a naive question: do I need to build two AppImages? One w/ libfuse2 and one w/ libfuse3 or is it transparent?

Samueru-sama:
> Hello, Just a naive question: do I need to build two AppImages? One w/ libfuse2 and one w/ libfuse3 or is it transparent?

No there is no need to do so. The appimage uses the static runtime, it works regardless if libfuse3 or libfuse2 is on the system. 

I'm sorry because I just noticed that the way I worded the title of the PR makes it seem like it has a dependency to libfuse3 when that is actually no the case.

Syllo:
I tested it and seems to work fine, thanks!

PR #303, opened 2024-06-28
Add Support for the Intel Xe driver & HWMON power values

Steve-Tech:
Hi, this PR adds basic support for Intel's new Xe driver, it doesn't support GPU utilisation (the fdinfo is different) but all other features that were working for i915 now work for Xe (on an A770).

The added power fields from the HWMON patch were also implemented, but only `power1_max` is present on both i915 and Xe. Temperature, fans, etc. were never implemented by Intel in the HWMON patch.

Also, the memory clock speed attributes (`mem_cur_freq_mhz` & `mem_max_freq_mhz`) don't seem to exist (in either drivers), should they be removed?

Thanks,
Steve

### DIFF
--- a/src/extract_gpuinfo_intel.c
+++ b/src/extract_gpuinfo_intel.c
@@ -354,33 +354,33 @@ void gpuinfo_intel_refresh_dynamic_info(struct gpu_info *_gpu_info) {
 
   RESET_ALL(dynamic_info->valid);
 
-  nvtop_device *intel_dev_auto = gpu_info->driver == DRIVER_XE ? gpu_info->driver_device : gpu_info->card_device;
+  // We are creating new devices because the device_get_sysattr_value caches its querries
+  const char *syspath;
+  nvtop_device *card_dev_noncached = NULL;
+  if (nvtop_device_get_syspath(gpu_info->card_device, &syspath) >= 0)
+    nvtop_device_new_from_syspath(&card_dev_noncached, syspath);
+  nvtop_device *driver_dev_noncached = NULL;
+  if (nvtop_device_get_syspath(gpu_info->driver_device, &syspath) >= 0)
+    nvtop_device_new_from_syspath(&driver_dev_noncached, syspath);
+  nvtop_device *hwmon_dev_noncached = NULL;
+  if (gpu_info->hwmon_device) {
+    if (nvtop_device_get_syspath(gpu_info->hwmon_device, &syspath) >= 0)
+      nvtop_device_new_from_syspath(&hwmon_dev_noncached, syspath);
+  }
 
+  nvtop_device *clock_device = gpu_info->driver == DRIVER_XE ? driver_dev_noncached : card_dev_noncached;
   // GPU clock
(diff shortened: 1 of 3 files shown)


cyear:
Hello, how can I see the temperature and power consumption?
image

src/extract_gpuinfo_intel.c
```diff
@@ -350,42 +354,41 @@ void gpuinfo_intel_refresh_dynamic_info(struct gpu_info *_gpu_info) {
 
   RESET_ALL(dynamic_info->valid);
 
-  nvtop_device *card_dev_copy;
-  const char *syspath;
-  nvtop_device_get_syspath(gpu_info->card_device, &syspath);
-  nvtop_device_new_from_syspath(&card_dev_copy, syspath);
+  nvtop_device *intel_dev_auto = gpu_info->driver == DRIVER_XE ? gpu_info->driver_device : gpu_info->card_device;
```
Syllo, line 357: I'll add a patch to re-introduce the copy. The problem is that libsystemd and libudev may/will cache these values so we need to get create a new device and query the sysattr to get up-to-date values ([see libsystemd implementation](https://github.com/systemd/systemd/blob/3f24fa57df552accc2a6f9ab4d36724ba7227eff/src/libsystemd/sd-device/sd-device.c#L2309))

Syllo: I guess calling `sd_device_set_sysattr_value` with NULL should also do the trick to clear the cache

Syllo:
Thanks for your patch!

I've reintroduced the temporary devices to avoid cached queries from the libsystemd/libudev.
Your PCIe speed seemed wrong so I switched back the device to the card and not the driver (seems to be what works better for other drivers too). Does that show better PCIe speeds?

Could you please test with my changes?

Steve-Tech:
Hi @cyear,

> Hello, how can I see the temperature and power consumption?

Temperature isn't exposed by the driver so it wouldn't be easy to show it, and power consumption doesn't seem to be supported on consumer cards; it does however measure total energy consumption in joules, which could possibly be converted into watts.

Steve-Tech:
Hi @Syllo,

> Thanks for your patch!

No worries!

> I've reintroduced the temporary devices to avoid cached queries from the libsystemd/libudev.

Cool thanks, I was confused why they were there.

> Your PCIe speed seemed wrong so I switched back the device to the card and not the driver (seems to be what works better for other drivers too). Does that show better PCIe speeds?

Nah, there's a [bug with the driver/card](https://community.intel.com/t5/Graphics/Intel-Arc-A770-PCIe-Speed-2-5GT-s-Width-x1/m-p/1455448) where it only reports PCIe x1, even though it does operate at the full bandwidth available. `lspci` also shows it as x1. By switching it back to the card it just states `N/A` as the PCIe speed, I was hoping the later cards would have this fixed, and show the correct speed.

> Could you please test with my changes?

Apart from the PCIe speed, it works fine! While I'm booted with Xe again, I'll try and figure out the fdinfo hopefully in the next few hours.

PR #324, opened 2024-11-01
Make appimage be able to work on any linux system (and some refactor of the script).

Samueru-sama:
These changes make the appimage be able to work on any linux system, be it 15 years old or even musl distros. this is done by bundling all the libraries the nvtop binary needs + the ld-linux.so. 

These changes are similar what the deploy everything mode of [go-appimage](https://github.com/probonopd/go-appimage/blob/master/src/appimagetool/README.md) does but I find it far easier to do it manually instead of fighting with the go-appimage parameters though. 

Tested on Artix, Alpine (musl) and Ubuntu 20.04 (old Glibc). 

The only downside is that this makes the appimage 2.4 MiB.

Also updated the readme to use 24.04 instead of 18.04, because with older versions you had to do stuff like using pip to install a newer cmake, none of this is needed anymore by building the appimage this way. 

Edit: I'm linking an artifact if you want to test it, remove the `.txt` from the end. 

[nvtop-3.1.0-x86_64.AppImage.txt](https://github.com/user-attachments/files/17599358/nvtop-3.1.0-x86_64.AppImage.txt)

### DIFF
--- a/AppImage/README.md
+++ b/AppImage/README.md
@@ -2,7 +2,7 @@
 
 ```bash
 podman pull ubuntu:18.04
-podman run --interactive --tty --rm --volume $PWD:/nvtop ubuntu:18.04
+podman run --interactive --tty --rm --volume $PWD:/nvtop ubuntu:24.04
 cd nvtop
 ./AppImage/make_appimage.sh
 ```
--- a/AppImage/make_appimage.sh
+++ b/AppImage/make_appimage.sh
@@ -1,54 +1,63 @@
 #!/usr/bin/env bash
 
+export ARCH="$(uname -m)"
+export APPIMAGE_EXTRACT_AND_RUN=1
+APPIMAGETOOL="https://github.com/AppImage/appimagetool/releases/download/continuous/appimagetool-$ARCH.AppImage"
+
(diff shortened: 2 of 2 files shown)


Syllo:
Hello,

Thank you for the patch.

Right, this bundles all the libraries and the dynamic linker in the Appimage.
I can't seem to find if it's recommended or anything online but I guess it's better for distro using an older glibc or no glibc

I can probably try a build on a distro with musl and link everything statically to reduce the size.

Samueru-sama:
> Hello,
> 
> Thank you for the patch.
> 
> Right, this bundles all the libraries and the dynamic linker in the Appimage. I can't seem to find if it's recommended or anything online but I guess it's better for distro using an older glibc or no glibc

It is fairly new practice done in [go-appimage](https://github.com/probonopd/go-appimage)

[Some examples here](https://github.com/AlexanderP/tesseract-appimage/issues/3)

> I can probably try a build on a distro with musl and link everything statically to reduce the size.

If you can build nvtop statically then yeah all of this can be avoided.

Syllo:
Sounds good, cheers.

PR #330, opened 2024-11-19
Fix compilation on FreeBSD and avoid crash while exiting (no hwmonDevice)

ocochard:
Tests report:
- amdgpu: working
- Intel: not working (gpu not detected)
- Nvidia: Unknown

### DIFF
--- a/include/ascend/dcmi_interface_api.h
+++ b/include/ascend/dcmi_interface_api.h
@@ -21,7 +21,7 @@ extern "C" {
 #endif
 #endif /* __cplusplus */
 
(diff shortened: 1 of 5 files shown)


Syllo:
Thanks for the patch!
I modified the patch slightly to avoid the crash in `nvtop_device_unref` function. I guess that the systemd version gracefully handles nullptr arguments but not udev.

VVD:
This software are in FreeBSD ports now: https://www.freshports.org/sysutils/nvtop/

PR #331, opened 2024-11-20
Intel additional HWMON (Fan + Temp) & i915 process memory usage

Steve-Tech:
Hi,

i915 has recently added fan speeds in kernel 6.12, this PR implements that ([patch](https://patchwork.freedesktop.org/patch/610458/)). There's also a temperature patch being developed upstream that I've also implemented support for ([patch series](https://patchwork.freedesktop.org/series/137874/)).

I don't know if it was always like this or recently implemented, but i915 also has `drm-total-local0` in the fdinfo, which is the GPU's memory used by that process. I also fixed a logic error that resulted in Xe reading the wrong value as memory usage.

The energy consumption was exposed by i915 in joules, I've now converted that into watts for nvtop to display.

Here's an example of my GPU:

nvtop example UI for an A770

Also it seemed like `hide_processes_list` was left uninitialized instead of false, causing it to randomly enable in my builds, which is also fixed.

I'm also working on using the drm ioctls to get the memory usage for i915, in my [intel-ioctl branch](https://github.com/Steve-Tech/nvtop/tree/intel-ioctl), but I'm omitted it from this PR since I didn't want to make it too big to review.

Thanks,
Steve

### DIFF
--- a/include/nvtop/extract_gpuinfo_common.h
+++ b/include/nvtop/extract_gpuinfo_common.h
@@ -100,6 +100,7 @@ enum gpuinfo_dynamic_info_valid {
   gpuinfo_pcie_rx_valid,
   gpuinfo_pcie_tx_valid,
   gpuinfo_fan_speed_valid,
+  gpuinfo_fan_rpm_valid,
   gpuinfo_gpu_temp_valid,
   gpuinfo_power_draw_valid,
   gpuinfo_power_draw_max_valid,
@@ -124,6 +125,7 @@ struct gpuinfo_dynamic_info {
   unsigned int pcie_rx;             // PCIe throughput in KB/s
   unsigned int pcie_tx;             // PCIe throughput in KB/s
   unsigned int fan_speed;           // Fan speed percentage
+  unsigned int fan_rpm;             // Fan speed RPM
   unsigned int gpu_temp;            // GPU temperature °celsius
   unsigned int power_draw;          // Power usage in milliwatts
   unsigned int power_draw_max;      // Max power usage in milliwatts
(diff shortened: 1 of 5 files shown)


Steve-Tech:
Hi @fkobi,

> I have an iris-xe and temperature does not show there, so there's progress I think.

The [temperature patch](https://patchwork.freedesktop.org/series/137874/#rev3) isn't mainlined yet, so you'd have to apply it manually. Otherwise maybe integrated graphics doesn't do temperature?

> However with `intel_gpu_top` I can see the clock speed and power consumption.

Oh okay, I'll look into this.

> also, looking at your screenshot should power be ` N/A  / 190 W`?

Yeah, the power consumption isn't reported in hwmon, but the max is.

Steve-Tech:
Hi @fkobi, power consumption is now calculated in this PR.
I also looked into `intel_gpu_top` and ended up exploring the ioctls supported by the i915 drm driver, where I found out the memory usage can be queried and is implemented (only for i915 atm) in my [intel-ioctl branch](https://github.com/Steve-Tech/nvtop/tree/intel-ioctl).

Steve-Tech:
Sorry @fkobi,

> I have an iris-xe and temperature does not show there, so there's progress I think.

I only just realised, temp, power, energy, and fan speed, are only available on the discrete DG1 and DG2 cards. It doesn't seem to be available for integrated graphics.

https://github.com/torvalds/linux/blob/7503345ac5f5e82fd9a36d6e6b447c016376403a/drivers/gpu/drm/i915/i915_hwmon.c#L847-L866

Syllo:
Hello,
Thanks for this patch.

I added a commit to display the RPM instead of guessing a maximum for the fan percentage value.
I assumed that `hwmon_fan` is the actual RPM (no post-processing needed). Does it look good with my commit on your end?

Steve-Tech:
Hi @Syllo,
No worries, and displaying the RPM is perfect!

I'll probably be creating another PR with memory usage sometime tomorrow (or whenever this gets merged).

Thanks for your work!
Steve

PR #339, opened 2024-12-09
Split differing parts of i915 and xe into seperate files, and implement ioctls for memory usage

Steve-Tech:
Hi,

Having both i915 and xe in the same file was getting annoying to maintain, so I've split the differing parts, and kept the similar parts in `extract_gpuinfo_intel.c`.
I've implemented the ioctls that mesa uses to get memory usage, so we can also get the memory usage (for both i915 and xe).
There's some minor additions like xe encoder usage, adding compute usage to the total usage, and showing gtt memory usage for integrated gpus.
The actual memory usage requires the CAP_PERFMON or CAP_SYS_ADMIN capability, (use `sudo setcap cap_perfmon=ep nvtop` to add it), but I've also changed the interface to still display the total memory (e.g. `N/A/15.906Gi`) if the memory usage is denied.

Here's an example of my GPU:
image

Thanks,
Steve


src/extract_gpuinfo_intel_i915.c
```diff
@@ -0,0 +1,295 @@
...
+#include "nvtop/device_discovery.h"
+#include "nvtop/extract_gpuinfo_common.h"
+#include "nvtop/extract_processinfo_fdinfo.h"
+#include "nvtop/time.h"
+
+#include "extract_gpuinfo_intel.h"
+
+#include <assert.h>
+#include <drm/drm.h>
+#include <drm/i915_drm.h>
```
Steve-Tech, line 31: This would require the linux userspace headers to be installed (`linux-libc-dev`), is that okay?

Syllo: It's the same requirement as for AMD's `#include <libdrm/amdgpu.h>` we need for AMDGPU so nothing new as far as I know :wink:

src/extract_gpuinfo_intel_xe.c
```diff
@@ -0,0 +1,264 @@
...
+  static char *line = NULL;
+  static size_t line_buf_size = 0;
+  ssize_t count = 0;
+
+  bool client_id_set = false;
+  unsigned cid;
+  nvtop_time current_time;
+  nvtop_get_current_time(&current_time);
+
+  union intel_cycles gpu_cycles = {.array = {0, 0, 0, 0, 0}};
```
Syllo, line 140: If you want to zero-initialize you can also use `= { 0 }` ( or `= {}` if the project uses the C standard >= C23)

src/extract_gpuinfo_intel_i915.c
```diff
@@ -0,0 +1,295 @@
...
+  struct gpuinfo_dynamic_info *dynamic_info = &gpu_info->base.dynamic_info;
+
+  if (gpu_info->card_fd) {
+    int32_t length = 0;
+    struct drm_i915_query_memory_regions *regions =
+        intel_i915_query_alloc(gpu_info->card_fd, DRM_I915_QUERY_MEMORY_REGIONS, &length);
+    if (regions) {
+      for (unsigned i = 0; i < regions->num_regions; i++) {
+        struct drm_i915_memory_region_info mr = regions->regions[i];
+        if (mr.region.memory_class == I915_MEMORY_CLASS_DEVICE || regions->num_regions == 1) {
```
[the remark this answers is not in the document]
Steve-Tech: Yeah that's correct, my ARC seems to have 2 memory regions, one device and one system, while UHD integrated graphics (not sure about Iris graphics but I assume it's the same) only has one for system memory.

src/extract_gpuinfo_intel_i915.c
```diff
@@ -0,0 +1,295 @@
...
+  if (ret != 0)
+    return -errno;
+  else if (item.length < 0)
+    return item.length;
+
+  *buffer_len = item.length;
+  return 0;
+}
+
+static inline void *intel_i915_query_alloc(int fd, uint64_t query_id, int32_t *query_length) {
```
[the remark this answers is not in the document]
Steve-Tech: Yeah sure, that was just a straight copy from the mesa source.

src/extract_gpuinfo_intel_xe.c
Steve-Tech: Ahh cool, thank you!

Syllo:
I had to add `xe_drm.h` to nvtop temporarily until libdrm catches up!
It shouldn't be a big deal since the file is backwards compatible by the fact that it's a user-space kernel interface!

oxalica:
Hi, I'm testing the master branch (c79a9c76fa2903d9ee2922d76848db0a5d9ff6b9) with this PR included. The total GPU memory is indeed available (with `sudo`), but the per-process GPU memory is still shown as N/A. Is this intended? I'm on NixOS unstable with Intel ARC A380.

image

Steve-Tech:
Hi @oxalica,

What kernel are you running? The required fdinfo key was added somewhat recently and I haven't yet gone though to figure out when.

oxalica:
> What kernel are you running? 
> Edit: Looks like it was added in [9688530](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=968853033d8aa4dbb80fbafa6f5d9b6a0ea21272) in kernel 6.6.
> Also you can see every process's drm fdinfo using `grep ^drm /proc/*/fdinfo/*`, there should be `drm-total-local0` if your kernel supports it, but let me know if the A380 is `drm-total-local1` or something.

Okay, I was using Linux 6.6.69 (LTS). I just tested 6.12.8: GPU memory is correctly shown and `drm-total-local0` is available. So it requires a newer kernel than 6.6 but not sure what exact version is it.

Anyway, thanks for the info.

Steve-Tech:
> So it requires a newer kernel than 6.6 but not sure what exact version is it.

Yeah sorry I was incorrect, the commit was based off 6.6-rc7, but it didn't get merged until v6.8-rc1.

PR #344, opened 2024-12-14
Workaround broken PCIe speeds on Intel Arc

Steve-Tech:
Hi again,

Intel Arc Alchemist has a weird (hardware?) bug, where the card and one of it's two PCIe bridges report their link speed as PCIe 1.0 x1, this PR implements a workaround to get the correct speed from it's first PCIe bridge.

e.g. `lspci -tv`:
```
-[0000:00]-+-01.1-[01-04]----00.0-[02-04]--+-01.0-[03]----00.0  Intel Corporation DG2 [Arc A770]
                                           \-04.0-[04]----00.0  Intel Corporation DG2 Audio Controller
             ^ Correct       ^ Incorrect     ^ Incorrect
```

I have checked a few probes on the [Linux Hardware Database](https://linux-hardware.org/) to verify whether it's just my system or card, and it's not. There is also a [bug report](https://community.intel.com/t5/Graphics/Intel-Arc-A770-PCIe-Speed-2-5GT-s-Width-x1/m-p/1455448) on Intel's website with some info.

I have no idea if this is fixed in Battlemage.

I've also added Intel to the libdrm detection in `src/CMakeLists.txt`, and moved AMDGPU and MSM support checks out of the libdrm detection block since `FATAL_ERROR` will cause CMake will exit anyway. I hope this is okay.

Thanks,
Steve

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -74,26 +74,27 @@ if(AMDGPU_SUPPORT OR INTEL_SUPPORT OR V3D_SUPPORT)
   target_sources(nvtop PRIVATE device_discovery_linux.c)
 endif()
 
-if(AMDGPU_SUPPORT OR MSM_SUPPORT OR PANFROST_SUPPORT OR PANTHOR_SUPPORT)
+if(AMDGPU_SUPPORT OR INTEL_SUPPORT OR MSM_SUPPORT OR PANFROST_SUPPORT OR PANTHOR_SUPPORT)
   # Search for libdrm for AMDGPU support
   find_package(Libdrm)
 
   if(Libdrm_FOUND)
     message(STATUS "Found libdrm; Enabling support")
     target_include_directories(nvtop PRIVATE ${Libdrm_INCLUDE_DIRS})
-    if (AMDGPU_SUPPORT)
-      target_sources(nvtop PRIVATE extract_gpuinfo_amdgpu.c)
-      target_sources(nvtop PRIVATE extract_gpuinfo_amdgpu_utils.c)
-    endif()
-
-    if (MSM_SUPPORT)
-      target_sources(nvtop PRIVATE extract_gpuinfo_msm.c)
-      target_sources(nvtop PRIVATE extract_gpuinfo_msm_utils.c)
-    endif()
-
   else()
-    message(FATAL_ERROR "libdrm not found; This library is required for AMDGPU and MSM support")
+    message(FATAL_ERROR "libdrm not found; This library is required for AMDGPU, INTEL, MSM, PANFROST and PANTHOR support")
+    # CMake will exit if libdrm is not found
   endif()
+endif()
+
+if (AMDGPU_SUPPORT)
+  target_sources(nvtop PRIVATE extract_gpuinfo_amdgpu.c)
+  target_sources(nvtop PRIVATE extract_gpuinfo_amdgpu_utils.c)
+endif()
+
+if (MSM_SUPPORT)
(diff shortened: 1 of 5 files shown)


Syllo:
Sounds reasonable, but I'd rather only enable this for the Alchemist devices (ARC) devices for now.

Looking at [Intel's hardware table](https://dgpu-docs.intel.com/devices/hardware-table.html) it seems that all the ARC PCI IDs starts with "0x56" so doing a check like `if (dev_id & 0xff00 == 0x5600)` before entering your workaround would be better imho.

Steve-Tech:
Yeah good point, it also looks like the PCI bridges are usually '4fa0' and '4fa4', so I can check those too. I'll definitely implement this sometime after next week!

Steve-Tech:
I found a [probe of an ARC B580](https://linux-hardware.org/?probe=aa2af1b2d7&log=lspci_all), and it seems like it has the same PCI topology and issue as the Alchemist cards; meaning this is not fixed in Battlemage.

stolk:
Thank you for making this fix.
I was wondering why nvtop would display something different from what lspci -vvv gave me.

But I still wonder why my Gen5 slot and Gen5 Arc B70 are still used as Gen4, though.
In my BIOS I changed the slot from Auto to Gen5, but still get:

```
$ sudo lspci -tv
-[0000:00]-+-00.0  Intel Corporation Raptor Lake-S 8+12 - Host Bridge/DRAM Controller
           +-01.0-[01-04]----00.0-[02-04]--+-01.0-[03]----00.0  Intel Corporation Battlemage G31 [Intel Graphics]
           |                               \-02.0-[04]----00.0  Intel Corporation Device e2f7
```

And the link status gives:

```
$ sudo lspci -s 01:00 -vvv | grep LnkSta:
		LnkSta:	Speed 16GT/s, Width x16
$ sudo lspci -s 03:00 -vvv | grep LnkSta:
		LnkSta:	Speed 2.5GT/s, Width x1
```

LnkCap on the bridge is reported as Gen4, which does not match my BIOS setting. Maybe the card failed to negotiate the Gen5?

Steve-Tech:
> Thank you for making this fix.

No worries!

> Maybe the card failed to negotiate the Gen5?

I don't know for certain, but I believe LnkCap should still be 32.0GT/s if the negotiation failed. Does `sudo dmidecode -t slot` list the slot as PCIe 5? Also, I'm sure you've checked this, but are you sure the GPU is in the Gen 5 supported slot?

stolk:
It is in the correct slot (closest to CPU, per manual) and dmidecode reports that slot as Gen5.

```
Handle 0x0028, DMI type 9, 17 bytes
System Slot Information
	Designation: PCIEX16(G5)
	Type: PCI Express
	Data Bus Width: 16x or x16
	Current Usage: In Use
	Length: Long
	ID: 0
	Characteristics:
		3.3 V is provided
		Opening is shared
		PME signal is supported
	Bus Address: 0000:01:01.0
```

It may be caused by the three M2 slots that I use on my board. Apparently, motherboards try to lower PCIe speeds based on what the M2 drives are using?

Steve-Tech:
> It may be caused by the three M2 slots that I use on my board. Apparently, motherboards try to lower PCIe speeds based on what the M2 drives are using?

Yeah possibly, what was your motherboard?

Interestingly, my `dmidecode` has `Type: PCI Express 4 x16` (it's not a gen 5 board) while yours just has `Type: PCI Express`. I guess maybe that field is just a vendor provided string haha.

Issue #343, opened 2024-12-13
DRM-related installation issue


georgebisbas:
Dear all, Hi!

First of all, thank you for this great tool. I/we have been using it for years, and we love it!
We are using nvtop in our CI/CD and it started to break the last 2 days.
We start with a brief story of how the bug started appearing in our CI and finish with a reproducible dockerfile
. I hope it helps and can reproduce the error on your side!

nvtop installs successfully up to commit:
(https://github.com/Syllo/nvtop/commit/9b7819ed8c9baea5fe4ff3d8b76c50477249fd6f)

The installation bug started to appear after merging this commit:
(https://github.com/Syllo/nvtop/commit/eddc07da7ea4fcfc91660db9f59750829d94a27c)

The complete error trace was:
```
7.702 /app/nvtop/src/extract_gpuinfo_intel_xe.c:31:10: fatal error: drm/xe_drm.h: No such file or directory
7.702    31 | #include <drm/xe_drm.h>
7.702       |          ^~~~~~~~~~~~~~
7.702 compilation terminated.
7.702 make[2]: *** [src/CMakeFiles/nvtop.dir/build.make:370: src/CMakeFiles/nvtop.dir/extract_gpuinfo_intel_xe.c.o] Error 1
7.702 make[1]: *** [CMakeFiles/Makefile2:143: src/CMakeFiles/nvtop.dir/all] Error 2
7.703 make: *** [Makefile:136: all] Error 2
```

The same error keeps appearing until this commit:
(https://github.com/Syllo/nvtop/commit/41023f2ace1d7e541cb1e9ba6170867c7fffdf9f)

obviously changing after merging:
(https://github.com/Syllo/nvtop/commit/56b78f1e8edd152db86429cb3d7d4e8aa8839f0c)

```c
extract_gpuinfo_intel_i915.c:110:57: error: 'DRM_I915_QUERY_MEMORY_REGIONS'
```
The complete trace is:

```bash
7.247 /app/nvtop/src/extract_gpuinfo_intel_i915.c: In function 'gpuinfo_intel_i915_refresh_dynamic_info':
7.247 /app/nvtop/src/extract_gpuinfo_intel_i915.c:110:57: error: 'DRM_I915_QUERY_MEMORY_REGIONS' undeclared (first use in this function)
7.247   110 |         intel_i915_query_alloc_fetch(gpu_info->card_fd, DRM_I915_QUERY_MEMORY_REGIONS, &length);
[...]

PR #346, opened 2024-12-17
Add GitHub Workflow to test for successful compilation

Steve-Tech:
Hi,

I'm sorry I broke compiling on Ubuntu 22.04 previously (#343), I've implemented a GitHub workflow to compile for Ubuntu 20.04, 22.04, and 24.04, to prevent that from happening again.

An example run can be found [here](https://github.com/Steve-Tech/nvtop/actions/runs/12368592818).

Thanks,
Steve

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -1,4 +1,4 @@
-name: Compile
+name: Compile Ubuntu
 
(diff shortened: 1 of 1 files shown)


Syllo:
Cheers

PR #349, opened 2024-12-24
Fixed typo on cmake command. Added note about building for Apple Silicon

deanwampler:
The `cmake` command needs `../..`, not `..`.

I added a note about the `cmake` command for Apple Silicon.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -339,7 +339,7 @@ pixi global install nvtop
 ```bash
 git clone https://github.com/Syllo/nvtop.git
 mkdir -p nvtop/build && cd nvtop/build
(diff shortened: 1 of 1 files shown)


Syllo:
Not sure if I follow:

```bash
# This clones into the nvtop directory (by git's default cloning in a directory with the repo's name)
git clone https://github.com/Syllo/nvtop.git
# Create a build directory in nvtop and switch to it
mkdir -p nvtop/build && cd nvtop/build
```
So at that point, the directory should look something like:
```
ntop
├── CMakeLists.txt
└── build <- we are here
```

Hence, I don't see the point of the second parent directory to the `cmake` command since `CMakeList.txt` is in the direct parent.

I could perhaps update the documentation with more recent cmake commands like `cmake -B nvtop/build -S nvtop && cmake --build nvtop/build`

PR #351, opened 2025-01-02
Add missing va_end() call

peter15914:
```va_end()``` was not called if ```nvtop_device_get_sysattr_value()``` returned negative value.
I added call to ```va_end()```.


Syllo:
Cheers

Issue #352, opened 2025-01-14
No encoding/decoding info with AMD RX 7700 XT, Linux 6.12.8


davidebeatrici:
![](https://github.com/user-attachments/assets/d3da896a-78e3-4da2-80fa-baf244109838)

davidebeatrici:
Yep:

![](https://github.com/user-attachments/assets/e921099d-eb0f-412c-ae7c-4abfa02b5568)
![](https://github.com/user-attachments/assets/f019af62-da8d-4c38-8d3e-b9f4cef9ab48)

Steve-Tech:
Hi @davidebeatrici, I think I've found the issue, would you be able to test my [amd-enc branch](https://github.com/Steve-Tech/nvtop/tree/amd-enc)?

Steve-Tech:
Thanks for testing, sorry I don't know what the issue might be then.

davidebeatrici:
Please note that when forcing `ENC/DEC` to be shown, the percentage stays at 0%.

Steve-Tech:
Ahh cool, but still not for encoding?

Syllo:
I'm seeing something similar when using [sunshine](https://github.com/LizardByte/Sunshine) to screen-cast the screen to an Apple TV with an RX 6800XT.

I don't see any encoding reported by the AMDGPU driver, however it's using vaapi for sure since it's not being encoded on the CPU!

When using ffmpeg to encode a video I see the encoder utilization though!

I don't know, it may be an AMDGPU driver bug that does not report encoder usage in some cases?

PR #353, opened 2025-01-16
Fix shared encode/decode not being displayed on AMD

Steve-Tech:
Hi,

The UI only shows the decode usage when `encode_decode_shared` is true but amdgpu will only use the encode usage field when it is shared, causing nvtop to never show any usage. I've now made it save the usage to the decode usage field when it's shared.
#352 seems related, but seemingly not solved.

I also made the ENC/DEC field longer, so that there's a proper usage bar and `100%` doesn't look bad.
Before:
image
After:
image

Thanks,
Steve

### DIFF
--- a/include/nvtop/interface_internal_common.h
+++ b/include/nvtop/interface_internal_common.h
@@ -59,6 +59,7 @@ struct device_window {
   WINDOW *mem_util_no_enc_and_dec;
   WINDOW *encode_util;
   WINDOW *decode_util;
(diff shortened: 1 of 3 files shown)


Syllo:
Thank you!

I remember it used to advertise the shared decode/encode as decode but it could have changed with a kernel version or newer GPUs …
Adding both metrics will hopefully work for both cases 😅

PR #360, opened 2025-02-22
Adding TPU support via libtpuinfo

rdyro:
Adding TPU support through an unofficial [libtpuinfo](https://github.com/rdyro/libtpuinfo) built by rewriting [tpu-info](https://pypi.org/project/tpu-info/) in golang (to compile it into a dynamic library).

I added a cmake rule to detect the presence of the necessary [libtpuinfo.so](https://github.com/rdyro/libtpuinfo/releases) library and enable the TPU support. Otherwise, it's possible to build with -DTPU_SUPPORT=ON following the convention.

### DIFF
--- a/src/extract_gpuinfo_tpu.c
+++ b/src/extract_gpuinfo_tpu.c
@@ -1,0 +1,301 @@
+/*
+ *
+ * Copyright (C) 2025 Robert Dyro <robert.dyro@gmail.com>
+ *
+ * This file is part of Nvtop
+ *
+ * Nvtop is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Nvtop is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with nvtop.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "nvtop/extract_gpuinfo_common.h"
+#include "nvtop/time.h"
+
+#include <fcntl.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+#include <math.h>
+#include <dlfcn.h>
+#include <sys/time.h>
+
+struct gpu_info_tpu {
+  struct gpu_info base;
+  int device_id;
+};
+
+struct tpu_chip_usage_data {
+  char name[8];
+  int64_t device_id;
+  int64_t memory_usage;
+  int64_t total_memory;
+  double duty_cycle_pct;
+  int64_t pid;
+};
+
+static bool gpuinfo_tpu_init(void);
+static void gpuinfo_tpu_shutdown(void);
+static const char *gpuinfo_tpu_last_error_string(void);
+static bool gpuinfo_tpu_get_device_handles(struct list_head *devices, unsigned *count);
+static void gpuinfo_tpu_populate_static_info(struct gpu_info *_gpu_info);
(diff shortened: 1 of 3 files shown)


rdyro:
@Syllo Could you take a look when you get a chance?

src/extract_gpuinfo_tpu.c
```diff
@@ -0,0 +1,290 @@
+/*
+ *
+ * Copyright (C) 2022 Maxime Schmitt <maxime.schmitt91@gmail.com>
```
Syllo, line 3: I'm flattered but you can put your own copyright 😉

src/extract_gpuinfo_tpu.c
```diff
@@ -0,0 +1,290 @@
...
+bool gpuinfo_tpu_init(void) {
+  char* error_msg;
+  char symbol_name[64];
+  char library_path[256];
+  nvtop_get_current_time(&last_cache_refresh);
+  // invalidate cache by putting it in the past
+  last_cache_refresh = nvtop_substract_time(last_cache_refresh, (nvtop_time){10, 0}); 
+
+  // Load dynamic library symbols
+  snprintf(library_path, sizeof(library_path), "%s", libname);
```
Syllo, line 157: Why not use libbame directly in dlopen?

src/extract_gpuinfo_tpu.c
```diff
@@ -0,0 +1,290 @@
...
+  void *handle = dlopen(library_path, RTLD_LAZY);
+  if (!handle) {
+      error_msg = dlerror();
+      if (error_msg != NULL) fprintf(stderr, "TPU support error: %s\n", error_msg);
+      return false;
+  }
+  snprintf(symbol_name, sizeof(symbol_name), "%s", "tpu_chip_count");
```
Syllo, line 164: Same here, you may have left all the redundant snprintf

src/extract_gpuinfo_tpu.c
```diff
@@ -0,0 +1,290 @@
...
+  if (error_msg != NULL) goto dlsymb_cleanup;
+  snprintf(symbol_name, sizeof(symbol_name), "%s", "tpu_metrics");
+  _tpu_metrics = dlsym(handle, "tpu_metrics");
+  error_msg = dlerror();
+  if (error_msg != NULL) goto dlsymb_cleanup;
+
+  // Discover TPU devices
+  tpu_chip_count = _tpu_chip_count();
+  if (tpu_chip_count == 0) {
+    fprintf(stderr, "Found 0 TPU devices on the system.\n");
```
Syllo, line 180: The printfs usually do not coexist well with the ncurses interface, better to remove them or put them behind a debug flag

src/extract_gpuinfo_tpu.c
```diff
@@ -0,0 +1,290 @@
...
+
+bool refresh_tpu_cache(void) {
+  if (is_cache_valid()) return true;
+  nvtop_get_current_time(&last_cache_refresh);
+  if (tpu_chip_count <= 0) return false;
+  if (_tpu_pids(_pids, tpu_chip_count) != 0) {
+    reset_tpu_cache(false);
+    return false;
+  }
+  for (int i = 0; i < tpu_chip_count; i++) latest_chips_usage_data[i].pid = _pids[i];
```
Syllo, line 121: tpu_chip_count is int64_t so better use these everywhere as induction variables or move tpu_chip_count to using int (having more than 2billion TPUs on the same machine is unlikely, I guess 🙃)

Syllo (commented):
Sorry for the late review!

src/extract_gpuinfo_tpu.c
rdyro: Sounds good, I'll move to `int64_t`

src/extract_gpuinfo_tpu.c
rdyro: Added an NDEBUG guard

src/extract_gpuinfo_tpu.c
rdyro: Right :slightly_smiling_face:

rdyro:
Done, thanks for the comments!

PR #367, opened 2025-04-01
icon location should be share/pixmaps per the xdg .desktop spec

### LINKED ISSUE #212: [cosmetic] icons -> pixmaps
https://github.com/Syllo/nvtop/blob/92a3e442f7f0fa6e70bca73f475fe5a567303a5e/CMakeLists.txt#L90

icons path is used when the icon is included on inconset/theme. pixmaps usually is used when the program only used the icon in the .desktop launcher

greetings

feinedsquirrel:
fixes #212 

icon location for the nvtop.desktop file should be in /usr/share/pixmaps, per the spec:
https://specifications.freedesktop.org/icon-theme-spec/latest/#directory_layout

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,7 +142,7 @@ install(FILES
   RENAME nvtop.1)
 install(FILES
   "${CMAKE_CURRENT_SOURCE_DIR}/desktop/nvtop.svg"
-  DESTINATION share/icons
+  DESTINATION share/pixmaps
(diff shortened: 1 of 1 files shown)


feinedsquirrel:
Eh, hold off on this, I may have misunderstood the spec, and additionally, the source of behavior in a downstream package. I will mark this WIP until I verify my understanding.

michel-slm:
See #370 - it should be hicolor/icons/apps since it's an SVG

Syllo:
Sounds good, thanks for looking into it.
I merged the related fix

PR #369, opened 2025-04-05
Fix the installed filename of the appstream metainfo

michel-slm:
Per `appstream-util validate-relax`, the filename should match the application ID.


Syllo:
Cheers

PR #370, opened 2025-04-05
Fix icon path

michel-slm:
SVG icons should go to hicolor/scalable/apps


feinedsquirrel:
This is the "more correct" place to put the icon.

PR #376, opened 2025-04-22
Fix matrix duplicate OS

QuentiumYT:
Recently, ubuntu 20.04 has been dropped. When updating the CI workflow, you added `ubuntu-latest` which is the equivalent of `ubuntu-24.04` already.

See https://github.com/actions/runner-images#available-images

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
 
       matrix:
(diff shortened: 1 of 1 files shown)


Syllo:
Thanks for noticing that. I don't mind having the redundant check for now since we are not running an heavy CI. As a bonus we will automatically pick up the latest Ubuntu version when it lands

Issue #405, opened 2025-09-12
Insuficient device memory clock field width


KaeLL:
The field width of the memory clock seems to be insufficient, as shown below

<img width="581" height="49" alt="Image" src="https://github.com/user-attachments/assets/b9f77012-c02c-499e-8ecc-61d277341d01" />

Of course this is only a problem if the intent is to show the effective clock rate (which accounts for DDR and PAM4) rather than the real physical clock rate.

airvzxf:
Hi, @KaeLL.

I understand that this issue was closed with your pull request https://github.com/Syllo/nvtop/pull/406. If it is correct, could you close this issue?

PR #406, opened 2025-09-13
Add device_field for memory clock

KaeLL:
Currently, GPU clocks and GPU memory clocks have
the same same field width size in the interface,
which is presumed to be in the thousands of MHz.
However, memory clocks reported by the GPU can be
in the tens of thousands range, presumably to
account for memory features such as PAM4 (like on
the RTX 4090). This causes the GPU memory clock
field to be one byte short when 5 digit clocks
are reported, cutting the 'z' from MHz.
This commit fixes that by adding a new
device_field for the memory clock that's one char
longer than the device_field for the GPU clocks,
and makes the appropriate changes in usage and
calculations that rely on these values.

Related issue: #405


KaeLL:
Before fix
Screenshot From 2025-09-12 15-58-30

After the fix
Screenshot From 2025-09-12 16-37-44

PR #411, opened 2025-10-10
Add unified memory support for NVIDIA DGX Spark

sbhavani:
Detect and handle NVIDIA GPUs with unified memory architecture (e.g., GB10) that use system RAM. When `nvmlDeviceGetMemoryInfo` fails or returns zero memory, fall back to querying system memory via `sysinfo()` to report total, used, and free memory statistics.

This approach mirrors the existing unified memory handling for Apple M1/M2 GPUs and enables proper memory monitoring for NVIDIA GPUs with unified memory.


Syllo:
Nice, I saw the announcement of these systems. Thank you for adding the support.

Issue #399, opened 2025-08-07
Add support for AMD GTT Memory


DocMAX:
Can you please add a bar to nvtop to show GTT Memory usage (beside VRAM) on AMD GPUs?

thomas-0816:
Hello,
I'm testing v3.3.1 on Ubuntu 22.04 with Ryzen 7 PRO 7840U and GTT:

<img width="1919" height="481" alt="Image" src="https://github.com/user-attachments/assets/59a9da33-6a48-485f-8188-31ab7a1e18b9" />

MEM seems correct. but llama-server in the process list only has 3% in my test, should it be sth. like 19%?

PR #428, opened 2025-12-13
When running on an AMD APU add VRAM + GTT together

superm1:
The MEM display currently shows just VRAM (which is the carveout of system memory).  However amdgpu will use VRAM or GTT interchangeably on APUs.  That is this isn't really useful for most APU users.

Detect that an APU is in use and add the two together.

Closes: https://github.com/Syllo/nvtop/issues/399


Syllo:
Sounds helpful, thanks.

PR #451, opened 2026-03-21
Add support for Tenstorrent accelerators

OctopusET:
Tenstorrent AI accelerator backend (Blackhole, Wormhole, Grayskull) via tt-kmd

### Features
- Temperature
- Power
- Fan RPM
- Clock
- PCIe bandwidth
- Process listing

image

I used `extract_gpuinfo_` prefix to match with other but I checked that Rockchip NPU uses `extract_npuinfo_`, but TPU/Ascend/Enflame all use `extract_gpuinfo_` even those aren't actually GPUs. So, please decide this.

And I used <hello at example dot> format for copyrights, I would change to normal one if you want.3

I can modify the tt-kmd to show the RAM usages, but since one tenstorrent device (at least blackhole) can run only one thing at once (I might be wrong), it's useless.

So we won't need to add graph too.

> Tested on Blackhole p100a with tt-kmd 2.6.0.


lu-zero:
image

Syllo:
Thank you for adding the support for your accelerator!

PR #454, opened 2026-03-22
Fix: make --no-processes work with --snapshot and --loop

### LINKED ISSUE #442: `--no-processes` does not work with `--snapshot`  or `--loop`
The recent updates make it much easier integrate `nvtop` with other software. Thank you!

It would be great if the snapshot and loop modes supported the `--no-processes` flag.

As of `3.3.2` these commands still show the `processes` array:

```bash
nvtop -sP
```

```bash
nvtop --snapshot --no-processes
```

```bash
nvtop -lP
```

```bash
nvtop --loop --no-processes
```

yinghao-w:
Fixes #442 so snapshots can be specified with out process info. Regardless of `--no-processes` process info is still calculated to find other GPU info, but open to suggestions on that.

### DIFF
--- a/include/nvtop/interface.h
+++ b/include/nvtop/interface.h
@@ -55,7 +55,7 @@ int interface_update_interval(const struct nvtop_interface *interface);
 
 bool show_information_messages(unsigned num_messages, const char **messages);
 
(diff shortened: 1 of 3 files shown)


Syllo:
Cheers

PR #455, opened 2026-03-27
Radeon support

gusteivos:
Add support for the legacy `radeon` driver with a new backend.

Also removes old `radeon` related code from the `amdgpu` backend.

GPU usage estimation is not very accurate since it is based on a snapshot of the `GRBM_STATUS` register. This could be improved by following an approach similar to `radeontop`.

GPU name detection can be improved in the future by using `amdgpu_ids.h`.

Process listing is not implemented for this backend.

Tested on:
- XFX R9 270X 2GB (PITCAIRN_6810)
  https://www.techpowerup.com/gpu-specs/xfx-double-d-r9-270x.b2484
- Radeon HD 6570 512MB (TURKS_6759)

(both with overheating problems)

Captura de tela de 2026-03-27 13-42-39

Built on:
- Debian 13
- Ubuntu 25


Syllo:
Thank you.
Any support is better than nothing at all!
I had a laptop with Bulldozer APU and a  Radeon integrated GPU at some point and I completely understand the overheating issue :wink:

PR #460, opened 2026-04-09
add user-configurable plot line colors via setup menu

graysky2:
Add support for customizing the color of each active plot line.

Pressing Enter cycles through the available colors: Red, Cyan, Green, Yellow, Blue, Magenta, White. Colors are applied immediately at runtime without restart.

neat


Syllo:
Cheers

Issue #449, opened 2026-03-16
DGX Spark support worked in 3.3.0, has been broken since 3.3.1


raziel2001au:
With the DGX Spark, support was added in 3.3.0, everything worked, but as of 3.3.1 it:
* Detects used memory correctly
* Detects total VRAM incorrectly (VRAM should be total system ram - ram used by system processes, right now it always seems to use 121GB as available VRAM, which is not correct, as you'll see in the screenshots)
* The graph is no longer displayed correctly, with the memory bar always showing 0 % usage

Screenshot from my compile of the 3.3.0 tag:
<img width="974" height="481" alt="Image" src="https://github.com/user-attachments/assets/6767aab2-5936-4cd5-b96e-551ed3f97fa0" />

Screenshot from my compile of the 3.3.1 tag (but the same issues exist in 3.3.2 and main):
<img width="975" height="483" alt="Image" src="https://github.com/user-attachments/assets/6a72fc70-e4ea-49b6-893a-9bbc2fc6767b" />

These screenshots were taken one after the other during an AI workload where about 60GB was used by other system processes, you'll note in the newer version it incorrectly reports a total of 121GB of VRAM, which it should be closer to 64GB, since the VRAM and system ram is shared, so when system ram is consumed, total VRAM reduces. Also, the memory graph line is completely broken, showing 0% memory usage.

It was working perfectly in 3.3.0, but in 3.3.1 and newer, nvtop is completely unusable on the DGX Spark.

bernardogv:
Confirmed on independent DGX Spark hardware. I built four revisions from source and captured what each actually displays, along with the underlying NVML state.

### Reproduction across four revisions

Built from pinned commits in clean Ubuntu 24.04 containers (`-DNVIDIA_SUPPORT=ON`, AMDGPU/Intel off), run against the GPU with `--gpus all`:

| revision | commit | memory display |
|---|---|---|
| 3.3.0 | `15e6c810` | `MEM[0.000Gi/113.003Gi]` |
| 3.3.1 | `5e9d0456` | `MEM[N/A/121.690Gi]` |
| 3.3.2 (latest release) | `76890233` | `MEM[N/A/121.690Gi]` |
| master @ 2026-05-06 | `3d4a953d` | `MEM[\|\|8.828Gi/121.690Gi]` |

Ground truth from `/proc/meminfo` at the same moment:

```
MemTotal:        121.69 GiB
MemAvailable:    112.94 GiB
```

**Summary:**

[...]

PR #463, opened 2026-04-15
Fix NVML memory reporting regression on coherent UMA platforms (Fixes…

parallelArchitect:
… #449)

On GB10 / DGX Spark, nvmlDeviceGetMemoryInfo returns NVML_SUCCESS with total == system MemTotal (~121GB). This prevents has_unified_memory from being set, causing incorrect VRAM reporting and broken memory graph since 3.3.1.

Fix: detect UMA by comparing NVML total against /proc/meminfo MemTotal. If total >= 90% of system RAM, classify as unified memory and use MemAvailable instead of MemTotal for display.

Note: requires validation on GB10 / DGX Spark hardware. Author does not have access to a coherent UMA system.    

References

NVML API documentation on SOC/UMA behavior: https://docs.nvidia.com/deploy/nvml-api/nvml-api-reference.html
Community NVML shim for GB10 UMA: https://forums.developer.nvidia.com/t/nvml-support-for-dgx-spark-grace-blackwell-unified-memory-community-solution/358869
NVML memory fix at the shim layer: https://github.com/parallelArchitect/nvml-unified-shim
btop PR: https://github.com/aristocratos/btop/pull/1611
nvitop PR: https://github.com/XuehaiPan/nvitop/pull/208

### DIFF
--- a/src/extract_gpuinfo_nvidia.c
+++ b/src/extract_gpuinfo_nvidia.c
@@ -655,9 +655,24 @@ static void gpuinfo_nvidia_refresh_dynamic_info(struct gpu_info *_gpu_info) {
     memory_info.version = 2;
     last_nvml_return_status = nvmlDeviceGetMemoryInfo_v2(device, &memory_info);
     if (last_nvml_return_status == NVML_SUCCESS) {
-      // Check if this is a unified memory GPU (total == 0 indicates unified memory)
       got_meminfo = true;
-      if (memory_info.total == 0) {
+      // Detect coherent UMA platforms (e.g. GB10 Grace Blackwell):
+      // NVML returns NVML_SUCCESS but total == full system MemTotal.
+      // This is not usable VRAM — treat as unified memory and use MemAvailable.
+      unsigned long long sys_mem_total = 0;
+      FILE *mf = fopen("/proc/meminfo", "r");
+      if (mf) {
+        char ml[256];
+        while (fgets(ml, sizeof(ml), mf)) {
+          if (sscanf(ml, "MemTotal: %llu kB", &sys_mem_total) == 1) {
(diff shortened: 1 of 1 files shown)


parallelArchitect:
For anyone needing the NVML memory fix now while this PR is under review — the fix is available in this fork: https://github.com/parallelArchitect/nvml-unified-shim

Syllo:
Hey,

Sorry I merged #466 before coming to your patch and resolved the merge conflicts.

I'm a bit concerned about the workaround (>= 90% of RAM) as it could probably falsely detect as UMA many systems, e.g. 32GB of RAM and a dedicated GPU with 32GB of GDDR. Maybe for now the only system like this has 128GB of RAM (and discrete GPU afaik don't go that far) so we could differentiate on this metric.

I guess that we'll have to see how NVIDIA is going to expose this through their NVML library to be able to avoid this scenario.

parallelArchitect:
Valid concern on the threshold. The proper detection is via CUDA device attributes:

cudaDevAttrPageableMemoryAccessUsesHostPageTables — true on GB10
cudaDevAttrConcurrentManagedAccess — true on GB10

Both together identify hardware-coherent UMA without relying on memory size comparison. SM 12.1 check is also an option but that's too narrow.

The NVML path is the problem — nvmlDeviceGetMemoryInfo returns NVML_SUCCESS with total == system MemTotal on GB10, which is correct behavior for UMA but breaks the has_unified_memory detection. The fix should detect UMA via device attributes first, then use /proc/meminfo MemAvailable for display on those platforms.

I don't have direct GB10 access but azampatti and dustin1925 in the community do — happy to coordinate validation if you want to iterate on a cleaner fix.

PR #467, opened 2026-05-01
fix: add missing delwin() calls in free_device_windows

danbedford:
`shader_cores`, `l2_cache_size`, and `exec_engines` were allocated with `newwin()` in `alloc_device_window()` but never freed in `free_device_windows()`. This caused a memory leak every time device windows were torn down (e.g. on refresh or hot-unplug).

The function now frees all windows it allocates, preventing resource leaks in long-running sessions.


Syllo:
Thank you

PR #470, opened 2026-05-05
fix: correctly check nvidia MIG mode status for process utilization

Henry-ZHR:
On common GPUs, like RTX:

```text
$ nvidia-smi --query-gpu=mig.mode.current
mig.mode.current
[N/A]
```

Process utilization OK

On A100:

```text
$ nvidia-smi --query-gpu=mig.mode.current
mig.mode.current
Disabled
Disabled
...
```

But no process utilization available
screenshot

This PR should bring process utilization back when MIG disabled


Syllo:
Nice catch of the double negation, thank you!

PR #481, opened 2026-06-18
Add PCIe RX/TX load to the utilization charts

### LINKED ISSUE #268: [Feature Request] Graph for PCIe bus load
nvtop already displays current PCIe link speed as well as current RX and TX values. Wouldn't this be enough to also generate a bus load percentage as metric for graphs? This way it would be easier to spot bottlenecks related to data transfer.

Basavaraja-MS:
## Summary

Closes #268.

Adds PCIe **RX** and **TX** load as plottable metrics in the line charts, alongside the existing GPU and memory utilization graphs.

- The load is derived from the already-collected `pcie_rx` / `pcie_tx` counters (KB/s) and normalized against the link's theoretical peak bandwidth (`max_pcie_gen` × `max_pcie_link_width`), so it fits the charts' fixed 0–100% scale.
- Two new metrics, `pcieRxRate` and `pcieTxRate`, are selectable per GPU from the **F2 → Chart** setup window and persist in the configuration file.
- The normalization math lives in a dedicated, unit-tested translation unit (`src/pcie_utilization.c`) rather than inline in `interface.c`.

### Design note: max vs. current link
The metric is normalized against the **maximum** link bandwidth, not the currently negotiated link. GPUs down-train PCIe when idle (often to Gen1/x4), so normalizing against the live link would make tiny idle transfers look like high load and give a moving denominator. Using the max gives a stable "% of peak PCIe capability," matching the "spot the bottleneck" intent of #268. (The issue phrased it as "current link speed" — flagging this as a deliberate choice.)

### Per-lane bandwidth basis
Effective per-lane data rates after PCIe encoding overhead (8b/10b for gen 1/2, 128b/130b for gen 3–5, FLIT for gen 6): 250 / 500 / 984.6 / 1969 / 3938 / 7563 MB/s.

### Scope / limitations
- No layout or header changes — zero impact on users who don't enable the metrics.
- Only meaningful where the vendor backend reports both PCIe throughput **and** `max_pcie_gen`/`max_pcie_link_width` (NVIDIA, AMD, etc.); otherwise the line stays at 0, consistent with other unsupported metrics.
- Still bound by `MAX_LINES_PER_PLOT` (4 metrics per plot).

## Test plan

- [x] `cmake -B build -DBUILD_TESTING=ON && cmake --build build` — clean, no warnings.
- [x] `ctest` — 8/8 pass, including 4 new `PcieUtilization` cases (per-gen bandwidth, width scaling, unknown-gen/missing-field → 0, percent scaling + clamp to 100%).
- [x] Manual: ran `nvbandwidth -t 0 -t 1 -b 256 -i 6` in a loop on an H200 NVL (Gen5 x16); enabled both metrics via F2. The `GPU0 PCIe RX%` / `PCIe TX%` lines peaked at ~95–99% under load and returned to 0 when idle, matching `nvidia-smi dmon -s t` (~60 GB/s of the 63 GB/s max).
- [x] Config round-trips: `pcieRxRate` / `pcieTxRate` written to / read from the config file via F12.

_Screenshot of the chart with the PCIe lines._
nvtop-output

Made with [Cursor](https://cursor.com)

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -62,6 +62,10 @@ NVTOP Options and Interactive Commands
 NVTOP has a builtin setup utility that provides a way to specialize the interface to your needs.
 Simply press ``F2`` and select the options that are the best for you.
 
+In the ``Chart`` section you can choose which metrics are plotted, including GPU and memory
+utilization, temperature, power, clocks, and the **PCIe RX / TX load** (the receive and transmit
+throughput as a percentage of the maximum link bandwidth).
+
 ![NVTOP Setup Window](/screenshot/Nvtop-config.png)
 
 ### Saving Preferences
--- a/include/nvtop/interface_common.h
+++ b/include/nvtop/interface_common.h
@@ -33,6 +33,8 @@ enum plot_information {
   plot_gpu_clock_rate,
   plot_gpu_mem_clock_rate,
   plot_effective_load_rate,
+  plot_pcie_rx_rate,
+  plot_pcie_tx_rate,
   plot_information_count
 };
 
--- a/include/nvtop/pcie_utilization.h
+++ b/include/nvtop/pcie_utilization.h
@@ -1,0 +1,37 @@
+/*
+ *
+ * Copyright (C) 2024 Maxime Schmitt <maxime.schmitt91@gmail.com>
+ *
+ * This file is part of Nvtop.
+ *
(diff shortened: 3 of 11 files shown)


Syllo:
Thank you for your contribution

PR #482, opened 2026-06-18
Show peak PCIe RX/TX throughput in the device header

Basavaraja-MS:
## Summary

Tracks the maximum PCIe RX and TX throughput observed since launch and displays it next to the instantaneous values in the device header:

```
PCIe GEN 5@16x RX: 1.2 GiB/s (max 57.5 GiB/s) TX: 48.3 MiB/s (max 59.8 GiB/s)
```

This makes transient transfer peaks visible at a glance when diagnosing bus bottlenecks, without having to watch the live value.

- Adds `pcie_rx_peak` / `pcie_tx_peak` to `struct device_window` (zero-initialized via the existing `calloc`; they persist across terminal resize since the struct isn't reallocated).
- The RX/TX rendering is refactored into a single `print_pcie_throughput_with_peak()` helper used for both directions (removes the prior copy-paste).
- The peak resets only when nvtop restarts.

### Note: header width
The PCIe header field is widened (46 → 88 columns) to fit the extra value, which increases each device panel's width. On a single wide terminal this is fine; on narrow or dense multi-GPU layouts it reduces how many panels fit per row. This is why the change is split out from the chart feature (#268 / the companion PR). Happy to make it opt-in via a config toggle if preferred.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/include/nvtop/interface_internal_common.h
+++ b/include/nvtop/interface_internal_common.h
@@ -73,6 +73,8 @@ struct device_window {
   WINDOW *shader_cores;
   WINDOW *l2_cache_size;
   WINDOW *exec_engines;
+  unsigned pcie_rx_peak; // Peak PCIe RX throughput observed since launch (KB/s)
+  unsigned pcie_tx_peak; // Peak PCIe TX throughput observed since launch (KB/s)
   bool enc_was_visible;
   bool dec_was_visible;
   nvtop_time last_decode_seen;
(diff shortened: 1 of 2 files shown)


Syllo:
Hello Basavaraja,

Thank you for contributing!

I think that this can be useful, but not in its current form. Right now it keeps the maximum for the whole duration of nvtop, so you'd have to restart nvtop every time you'd like to get a meaningful maximum throughput. What would be more useful is a rolling maximum, say over a couple of seconds.

Syllo:
Hello Basavaraja,

Thank you for contributing!

I think that this can be useful, but not in its current form. Right now it keeps the maximum for the whole duration of nvtop, so you'd have to restart nvtop every time you'd like to get a meaningful maximum throughput. What would be more useful is a rolling maximum, say over a couple of seconds.

PR #485, opened 2026-06-23
docs: mention invalid TERM ncurses error 

raravind007:
What changed:
- Added a troubleshooting note for ncurses failures when $TERM is unset or invalid.

Why:
- nvtop may exit with "ncurses: cannot initialize terminal type" in sandboxed or minimal terminals when $TERM is unset or invalid.
- Setting $TERM to a valid terminal type, such as xterm-256color, resolves this.

Testing:
- Built nvtop locally on Fedora 44.
- Verified the error occurred with TERM=unknown and was resolved with TERM=xterm-256color locally.


Syllo:
Cheers

PR #489, opened 2026-07-07
interface: prevent divide-by-zero for missing power limit

JoseOliveira-beep:
amdgpu: fall back to power1_cap_default when power1_cap is zero


Syllo:
Cheers

PR #491, opened 2026-07-10
feat(ascend): support DDR memory and fix device enumeration for 310P3

yulink1981:
- Add DDR memory fallback when HBM is unavailable (310P3/300i DUO)
- Fix device enumeration: use physical card ID instead of array index
- Fix per-process GPU utilization garbage values by zeroing process structs
- Add memory_type static field to distinguish DDR vs HBM in display
- Update TUI to show memory type (DDR 451MHz vs HBM xxxMHz)
- Update JSON snapshot output with memory_type field


Syllo:
Cheers

PR #492, opened 2026-07-14
amdgpu: count GTT in per-process memory usage

rborkow:
## Problem

Per-process memory on amdgpu only reads the `drm-memory-vram` fdinfo key.
On integrated GPUs, nearly all allocations live in the GTT heap, so the
process list reports ~0 memory while the device-level metrics show
gigabytes in use. This resolves the existing TODO at the parse site
(`// TODO: do we count "gtt mem" too?`).

Measured on a Ryzen AI 9 HX 370 / Radeon 890M (gfx1150, kernel 6.18),
with a 24 GB LLM loaded by ollama's llama-server:

    $ grep drm-memory /proc/<pid>/fdinfo/7
    drm-memory-cpu:         0 KiB
    drm-memory-gtt:  25258000 KiB   <- the model
    drm-memory-vram:     5348 KiB   <- all nvtop counted

nvtop showed the device at 24.3/48.0 GiB used, but attributed only
5 MiB to the process — and `gpu_mem_usage` read 0%.

## Change

Count `drm-memory-gtt` (and the legacy `gtt mem` key) together with
VRAM in `parse_drm_fdinfo_amd`, accumulating both into
`gpu_memory_usage`. This matches the device-level accounting, which
already sums both heaps via `amdgpu_query_info` (`total_memory` /
`used_memory` = vram + gtt). With numerator and denominator now
measuring the same thing, `gpu_mem_usage` percentages become coherent:
the process above reports 24.1 GiB / 50%, agreeing with the device
total.

## Behavior change on discrete GPUs

A process's reported memory now includes its GTT usage (host RAM mapped
for the GPU) on dGPUs as well. I'd argue this is consistent — the
device-level metric already includes GTT, and both heaps are GPU-mapped
memory — but if you'd prefer to keep dGPU semantics unchanged, I'm happy
to rework this to add GTT only when `static_info->integrated_graphics`
is set.

## Testing

Verified on the Radeon 890M above (before: ~0 per process; after:
per-process totals match device totals under a 24 GB workload).

### DIFF
--- a/src/extract_gpuinfo_amdgpu.c
+++ b/src/extract_gpuinfo_amdgpu.c
@@ -786,6 +786,8 @@ static void gpuinfo_amdgpu_refresh_dynamic_info(struct gpu_info *_gpu_info) {
 static const char drm_amdgpu_pdev_old[] = "pdev";
 static const char drm_amdgpu_vram_old[] = "vram mem";
 static const char drm_amdgpu_vram[] = "drm-memory-vram";
+static const char drm_amdgpu_gtt_old[] = "gtt mem";
+static const char drm_amdgpu_gtt[] = "drm-memory-gtt";
 static const char drm_amdgpu_gfx_old[] = "gfx";
 static const char drm_amdgpu_gfx[] = "drm-engine-gfx";
 static const char drm_amdgpu_compute_old[] = "compute";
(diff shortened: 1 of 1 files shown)


Syllo:
Cheers

PR #494, opened 2026-07-24
Display GPU architecture in json

Mahsuba:
Knowing the GPU architecture (like "Blackwell", "Rubin" or "RDNA 4") is often more useful than just having the model identifier, as the same architecture can be reused across various models with vastly different names (but which still have similar behaviour).

This PR adds a device_architecture field to the printed json, with mappings to several common architecture names for both AMD and NVidia.


Syllo:
Cheers

PR #495, opened 2026-07-25
Minor fixes for Intel Xe cards + Memory frequency

### LINKED ISSUE #450: Intel iGPU, `xe` kernel driver: nvtop reports `cur_freq` (requested frequency) instead of `act_freq` (actual frequency)
For at least Intel iGPUs using the `xe` kernel driver, and likely the `i915` driver as well, `nvtop` reports the value from `cur_freq` as the actual GPU frequency when Intel uses this as the "requested frequency" and use `act_freq` as the "actual frequency" the GPU is currently operating at.

For better accuracy, it would be nice if `nvtop` could report `act_freq` for the GPU frequency with Intel GPUs.

I noticed this issue when comparing `nvtop` with `qmassa`.

Below is the `qmassa` issue where ulissesf (qmassa's developer) helped explain why I was seeing different values with each tool:

https://github.com/ulissesf/qmassa/issues/35#issuecomment-4090592954

### LINKED ISSUE #493: Intel Arc Pro B60: power*_crit is reported as max power while power*_cap is 200 W
On an Intel Arc Pro B60 using the `xe` driver, Mission Center reports 400 W as the maximum power value while the configured hwmon power cap is 200 W.

The card's hwmon device exposes:

```text
power1_label=card
power1_cap=200000000
power1_crit=400000000
power1_max is absent
```

Linux hwmon documents `power*_cap` as the limit above which the system should reduce power use. It documents `power*_crit` as a critical maximum threshold that may trigger drastic action.

The Intel code in NVTOP commit `3d4a953da02bc18886734613bb9f60ff80669de7` falls back from `power*_max` to `power*_crit` for Battlemage and stores that value as `power_draw_max`. Mission Center 1.1.0 embeds this revision and displays the resulting 400 W value as the maximum.

Is `power*_crit` intended to represent the user-visible maximum in this case, or should NVTOP expose the configured power cap separately when both values differ?

Related change: <https://github.com/Syllo/nvtop/pull/393>

System details:

```text
GPU: Intel Arc Pro B60 [8086:e211]
Driver: xe
Kernel: 7.1.3-ogc3.4.fc44.x86_64
Mission Center: 1.1.0 Flatpak
Embedded NVTOP commit: 3d4a953da02bc18886734613bb9f60ff80669de7
```

This was reproduced through Mission Center's embedded NVTOP backend. I have not reproduced it with a standalone `nvtop` binary.

Kernel hwmon documentation: <https://docs.kernel.org/hwmon/sysfs-interface.html>

Steve-Tech:
Hi, this PR makes the following changes:

- Fixes #493
- Add max GPU memory frequency
    - For i915, this only seems to appear in the backport drivers: https://github.com/intel-gpu/intel-gpu-i915-backports/blob/88e15eff7c9ce9eb3941333088d38cf571960cfd/drivers/gpu/drm/i915/gt/intel_gt_sysfs_pm.c#L903-L935
    - For Xe, this only seems to be available for 'Ponte Vecchio' data center GPUs: https://lore.kernel.org/all/20240109110418.2065101-1-sujaritha.sundaresan@intel.com/
        - FWIW removing the `XE_PVC` check adds the files to sysfs, but they do not work on my A770 (`No such device or address`), probably due to lack of HBM.
- Partially Fix #450
    - Implement perf events, for more reliable measurement of the GPU frequency, however this now also requires `CAP_PERFMON`.
    - The underlying `cur_freq` vs `act_freq` is still present without `CAP_PERFMON`, however `act_freq` frequently returns `0`, so I've left `cur_freq` for now.

I have a question:
- Intel only exposes the max memory frequency, is it okay to use this for `mem_clock_speed` so that it gets displayed, or would you prefer I use `mem_clock_speed_max`?

Thanks,
Steve

### DIFF
--- a/src/extract_gpuinfo_intel.c
+++ b/src/extract_gpuinfo_intel.c
@@ -28,8 +28,10 @@
 
 #include <assert.h>
 #include <fcntl.h>
+#include <linux/perf_event.h>
 #include <stdio.h>
 #include <string.h>
+#include <sys/syscall.h>
 #include <unistd.h>
 #include <uthash.h>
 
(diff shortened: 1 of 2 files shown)


stolk:
Tested on Intel Arc B70, and works as expected. Max power now properly reported.

Syllo:
Looks good. I added a couple of small fixups.

I made it so that we display the max clock if we don't have the information of the current one. My rational is that displaying the max value is better than having no information on the frequency at all (`N/A`).

Syllo:
If that works for you, mark the PR as ready and I'll merge it

Steve-Tech:
Hi @Syllo, thank you for fixing this up! I meant to work on the frequency reporting a bit more, but got busy with other things.

I've tested everything on an A770 and B60, and it all works okay. The A770 in particular matches the GPU clock speed from `qmassa`. I'm ready for this to be merged too.

Syllo:
Cheers

PR #497, opened 2026-08-06
Display PCIe utilization on chart

destenson:
This adds the ability to display PCIe utilization history with a bar chart in the background of the regular utilization chart.

### DIFF
--- a/include/nvtop/interface_internal_common.h
+++ b/include/nvtop/interface_internal_common.h
@@ -140,6 +140,16 @@ struct nvtop_interface {
   unsigned num_plots;
   struct plot_window *plots;
   interface_ring_buffer saved_data_ring;
+  // PCIe throughput history, kept separately from saved_data_ring so it is not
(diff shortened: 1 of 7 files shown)


Syllo:
Thanks for this, @destenson.

Since this PR was opened, `master` gained the same capability via #481 ("Add PCIe RX/TX load to the utilization charts", merged 2026-09-07). That adds *PCIe RX load rate* / *PCIe TX load rate* as selectable chart metrics, normalizing the `pcie_rx`/`pcie_tx` counters against the link's maximum bandwidth through `pcie_utilization.c` (`pcie_max_bandwidth_kbs()` / `pcie_load_percent()`), and it is documented in the README/manpage and covered by unit tests.

So the feature this PR targets — displaying PCIe utilization on the chart — is already available, and I'm closing this one as superseded to avoid maintaining two implementations of the same link-bandwidth math.

The overlay here is a genuinely different presentation (always-on background shading rather than an opt-in line), so if you'd still like that, a much smaller follow-up built on `pcie_max_bandwidth_kbs()` would be welcome. Happy to reopen or help if so.

Thanks again for the work.

Issue #426, opened 2025-12-05
[Bug/Support] NVIDIA GB10 (Grace Blackwell) Reporting Issues (PCIe, Fan, Mem)


chromedot:
### _Generated with my good buddy Antigravity + Gemini 3 Pro High_

**Description
I am running nvtop on the new NVIDIA GB10 (Grace Blackwell) platform (Dell Pro Max Workstation Gb10 = Nvidia DGX Spark). While it correctly detects the GPU name and utilization, several metrics are reported incorrectly or as N/A, likely due to the SoC nature of the architecture (NVLink-C2C instead of traditional PCIe).**

System Information
Hardware: NVIDIA GB10 (Grace Blackwell Superchip)
Architecture: ARM64 (aarch64)
OS: Ubuntu 24.04.3 LTS
NVIDIA Driver: 580.95.05
CUDA Version: 13.0
Kernel: Linux 6.8.0-49-generic (aarch64)

Observed Issues
PCIe Misreporting:
Reported as: PCIe GEN 1@ 1x
Context: This is an SoC (System on Chip) where the GPU is connected via NVLink-C2C to the Grace CPU, not a standard PCIe slot. It seems nvtop is falling back to a default/placeholder value.
Memory Usage (Header):
Reported as: MEM[ N/A] in the bar gauge.
Context: nvidia-smi also reports "Not Supported" for memory usage on this platform (Unified Memory architecture). However, the process list at the bottom of nvtop does correctly show memory usage (e.g., 63931MiB for my python process). It would be great if the header gauge could aggregate the process memory or use the same source as the process list.
Fan Speed:
Reported as: FAN N/A%
Context: Likely managed by the chassis controller or SoC thermal subsystem, not exposed via standard NVML fan APIs for discrete cards.
Memory Clock:
Reported as: MEM N/A MHz
Working Metrics
GPU Utilization: Correctly shows load (e.g., 96%).
Power Usage: Correctly shows wattage (e.g., 58 W).
Process List: Correctly identifies processes and their per-process memory usage.
Additional Context 
**The GB10 uses the Unified Memory architecture (LPDDR5X) shared between CPU and GPU. The standard NVML queries for "FB Memory" might need to be adjusted or replaced with queries for the unified memory pool on this architecture.**

Output of nvidia-smi for reference:

+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.95.05              Driver Version: 580.95.05      CUDA Version: 13.0     |
[...]

PR #499, opened 2026-08-16
nvidia: don't report fake PCIe link on unified-memory SoC GPUs

Code4me2:
## Summary
Fixes the PCIe misreporting portion of **#426** on unified-memory SoC GPUs such as the DGX Spark (GB10).

On the GB10, the GPU is attached via NVLink-C2C (on-die interconnect), not a PCIe slot. However, NVML's `nvmlDeviceGetCurrPcieLinkGeneration` / `nvmlDeviceGetCurrPcieLinkWidth` return `NVML_SUCCESS` with placeholder values (`GEN 1 @ 1x`), which nvtop displayed as if they were a real PCIe link.

These devices are already detected (unified memory → `has_unified_memory`), so this PR gates PCIe link gen/width reporting on **not** being unified memory, making the header fall back to `N/A`. PCIe RX/TX throughput was already `N/A` here because NVML returns `NVML_ERROR_NOT_SUPPORTED` on this platform.

## Before (master)
```
Device 0 [NVIDIA GB10] PCIe GEN 1@ 1x RX: N/A TX: N/A
```

## After
```
Device 0 [NVIDIA GB10] PCIe GEN N/A RX: N/A TX: N/A
```

Memory (`MEM[...Gi]`) is already reported correctly on GB10 via the #466 unified-memory fallback; fan and memory-clock remain correctly `N/A` since the SoC does not expose them through NVML.

## Validation
- Built and tested **on a DGX Spark (GB10, aarch64, driver 580.173.02)**.
- Discrete (non-unified) GPUs are unaffected: `has_unified_memory` is only set when NVML reports unified/SoC memory.

### DIFF
--- a/src/extract_gpuinfo_nvidia.c
+++ b/src/extract_gpuinfo_nvidia.c
@@ -702,14 +702,20 @@ static void gpuinfo_nvidia_refresh_dynamic_info(struct gpu_info *_gpu_info) {
   }
 
   // Pcie generation used by the device
-  last_nvml_return_status = nvmlDeviceGetCurrPcieLinkGeneration(device, &dynamic_info->pcie_link_gen);
-  if (last_nvml_return_status == NVML_SUCCESS)
-    SET_VALID(gpuinfo_pcie_link_gen_valid, dynamic_info->valid);
+  // On unified-memory SoC platforms (e.g. DGX Spark / GB10) the device is not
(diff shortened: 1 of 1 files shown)


Syllo:
Cheers. I modified it slightly to query NVML for the BUS type instead, which I believe to be more robust than the unified memory flag

PR #515, opened 2026-09-19
fdinfo: apply the graphical default per process, not per fd

stolk:
A process can hold several DRM fds for the same GPU. The Intel xe and i915 callbacks only set a process type for an fd that has render or compute activity, so the idle fds came back as gpu_process_unknown and were defaulted to graphical.

This caused pure compute processes to show up erroneously as graphics+compute.

Merge the per-fd types first and only fall back to graphical when the merged type of the process is still unknown.


stolk:
Let me know if you want me to create an issue for this first.

But this fixes the incorrect G+C label for processes that are pure compute.

